PR #10, opened 2023-07-20
More values in channel info

theHenks:
Added string number, string position, HV card, DAQ card and CC4 card to the channel info creator to have more values accessible for sorting detectors. Especially useful if you want e.g. to investigate just one string, build visualization maps or look into specific electronic cards and their effects on the detectors. Screenshot shows example table:
image
@oschulz The naming of the columns in the table can changed. This is just my proposal.


oschulz:
LGTM.

PR #15, opened 2023-11-08
Change Usability Key

theHenks:
The usability key in the metadata has different meaning and treating it static as `"on" = true` is not enough since some channels have `no_psd` and `ac` and have to be processed anyhow. Changed this field to be a Symbol so that different selections can be performed. 
image


oschulz:
Can you have a look at the test? They should pass except on Windows (tests are currently broken on Win, we need to fix that separately).

theHenks:
I edited the tests. Should work now. You can now not check the usability anymore with `$usability` and you get a bool return. 
You have to explicitly state a usability state such as `:on` or `:no_psd` etc.
I think it is ready to merge.

PR #19, opened 2023-12-04
Include pulser and similar channels in channelinfo

oschulz:
CC @theHenks , can you give this a try?


theHenks:
@oschulz Seems fine to me. I see all aux channels, the pulser and the baseline trigger. Can be merged!

PR #23, opened 2024-02-02
Bug Fixes

theHenks:
- Fixed problems for `channelinfo` together with `RunCategorySelLike` for `channelinfo` with a single `DetectorId` as inpu
- Added `DataPartition` struct to be used for parititons and therefore the naming within all future partitioning functions
- Made `DataSelector` compatible with `PropDicts`. Otherwise, this doesn't work:
``` julia
dets = channelinfo(l200, (:p03, :r000, :cal); system=:geds, only_processable=true).detector
l200.par.rpars.ecal.p03.r000[dets[1]]
```

### DIFF
--- a/src/dataprod_config.jl
+++ b/src/dataprod_config.jl
@@ -103,15 +103,15 @@ export pydataprod_parameters
 
 
 """
-    data_partitions(data::LegendData, label::Symbol = :default)
+    partitioninfo(data::LegendData, label::Symbol = :default)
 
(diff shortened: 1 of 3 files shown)


theHenks:
@oschulz I didn't touch the `data_partitions` function yet since it has `IDicts` as keys so far. In my opinion and for consistency, the keys should be `DataPartitions`.

theHenks:
@oschulz PR is ready to merge from my side.

PR #27, opened 2024-02-06
Add LegendReport and ProcessStatus

oschulz:


oschulz:
@theHenks , you can give it a first try (see example in LegendReport docstring).

theHenks:
That is perfect for me. Just one small question:
What is the preferred way to write to Markdown. Something like that?
``` julia
open("report.md", "w") do io
    println(io, report)
end
```

oschulz:
To write Markdown to disk, that would be

```julia
open("report.md", "w") do io
    show(io, MIME"text/markdown"(), report)
end
```

but we can add a convenience function, of course.

PR #43, opened 2024-04-14
Add SLURM support

oschulz:
@theHenks


oschulz:
Usage example: Allocate SLURM resources, e.g. via

```shell
salloc --ntasks=4 --cpus-per-task=8 --mem-per-cpu=8GB --time=1:00:00
```

Then start Julia within the shell started by `salloc`:

```shell
julia
```

Then, in the Julia session, run

```julia
using LegendDataManagement

# Optional, workers should be shut down anyway, this just makes extra sure:
# LegendDataManagement.shutdown_workers_atexit()

legend_addprocs()
```

Query workers:

```julia
using TypedTables
Table(LegendDataManagement.distributed_resources())
```

`legend_addprocs()` should also work in a Julia process started via `sbatch`.

PR #49, opened 2024-06-14
Add plot recipes for parameter PropDicts

fhagemann:
The plot recipe takes three arguments:
- `chinfo`: the channel info with all detectors to be plotted on the x-axis
- `pars`: a `PropDict` that has the detector IDs as keys and parameters as values
- `properties`: an array of `Symbols` to access the data that should be plotted
  (if no `properties` are provided, the `PropDict` `pars` is expected to just contain the data to be plotted as values)

There are also keyword arguments:
- `calculate_mean`: If set to `true`, then the mean values are included in the legend labels. For values with uncertainties, the mean values are calculated as weighted means.
- `verbose`: some output when the plot is generated, e.g. if values for (some) detectors are missing

Examples:
```julia
using RecipesBase, Plots, LegendDataManagement
l200 = LegendData(:l200)
filekey = FileKey("l200-p03-r000-cal-20230311T235840Z")
#pars = l200.par.cal.p_energy(filekey);
pars = l200.par.ppars.ecal(filekey)
chinfo = channelinfo(l200, filekey; system = :geds, only_processable = true)
properties = [:e_cusp_ctc, :fwhm, :qbb];
plot(chinfo, pars, properties, verbose = true, color = 1, markershape = :o, calculate_mean = true)
```
image

```julia
using PropDicts
new_pars = PropDict(Dict(Symbol.(chinfo.detector) .=> chinfo.mass))
plot(chinfo, new_pars, color = 1, markershape = :x, verbose = true)
```
image


ext/LegendDataManagementRecipesBaseExt.jl
```diff
@@ -3,8 +3,135 @@
...
+    end
+    ylims := yl
+    @series begin
+        let (xticks, xlabels) = (eachindex(labels), labels)
+            label := ""
+            markeralpha := 0
+            y0 = @. zero(xticks) + yl[1] - 0.02*(yl[2] - yl[1])
+            # TODO: if text/font definitions get moved from Plots to RecipesBase,
+            # replace the following line with something Plots-independent
+            series_annotations := [Main.Plots.text(xlabels[i], 8, labelcolors[i], :right, rotation=90) for i in eachindex(xlabels)]
```
fhagemann, line 101: There is one method (`text`) in `Plots` which is not part of `RecipesBase`. Here, I used some trick to have no dependency on `Plots` (i.e. using `Main.Plots`) because `Plots` will always be loaded to the `Main` session to parse this plot recipe.

@oschulz Are you aware of a more elegant way to use this until the `text` and `font` methods are included/migrated to `RecipesBase`?

fhagemann:
The last force-push was just a rebase using the changes in #50

fhagemann:
The tests are still failing because the LegendTestData was not updated to the new format of saving `enrichment` in the metadata, which (instead of just having a value, now have the fields `val` and `unc` for value and uncertainty), see https://github.com/legend-exp/legend-detectors/commit/f87cf1c471439f0d7407670949a9f7da78adc009

The tests fail as `LegendDataManagement` was already updated to the new format (see https://github.com/legend-exp/LegendDataManagement.jl/commit/49da84425a3b5b93045670cf7757fd00b978b08c), but not the test data.

I opened a PR to legend-testdata (see https://github.com/legend-exp/legend-testdata/pull/16) to update the detector config files in the test data to this new format, but it might take some steps until they make it to the tests.

theHenks:
Test issue is due to wrong metadata in `LegendTestData`. This leads to failing tests.

PR #63, opened 2024-09-10
Active and total volume calculation

fhagemann:
`channelinfo` now has a new keyword: `detailed` (can also be renamed to something else) with default value `false`.

If set to `true`, the function `channelinfo` will calculate the total volume and the active volume of all `:geds` based on the FCCD values provided in the metadata and adds them as columns to the output of `channelinfo`.

There are warnings for:
- detectors with `cracks` in the metadata (the functions to calculate their volumes are still missing)
- detectors with no FCCD values provided (there, the active volume is equal to the total volume)

### DIFF
--- a/src/legend_data.jl
+++ b/src/legend_data.jl
@@ -238,7 +238,7 @@ const _cached_channelinfo = LRU{Tuple{UInt, AnyValiditySelection}, StructVector}
 Get all channel information for the given [`LegendData`](@ref) and
 [`ValiditySelection`](@ref).
 """
-function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol = :all, only_processable::Bool = false, detailed::Bool = false)
+function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol = :all, only_processable::Bool = false, extended::Bool = false)
     key = (objectid(data), sel)
     chinfo = get!(_cached_channelinfo, key) do
         chmap = data.metadata(sel).hardware.configuration.channelmaps
@@ -296,7 +296,7 @@ function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol
                 location, detstring, fiber, position
             )
 
-            if detailed
+            if extended
                 cc4::StaticString{8} = get(chmap[k].electronics.cc4, :id, "")
                 cc4ch::Int = get(chmap[k].electronics.cc4, :channel, -1)
                 daqcrate::Int = get(chmap[k].daq, :crate, -1)
--- a/src/legend_data.jl
+++ b/src/legend_data.jl
@@ -238,7 +238,7 @@ const _cached_channelinfo = LRU{Tuple{UInt, AnyValiditySelection}, StructVector}
 Get all channel information for the given [`LegendData`](@ref) and
 [`ValiditySelection`](@ref).
 """
-function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol = :all, only_processable::Bool = false)
+function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol = :all, only_processable::Bool = false, detailed::Bool = false)
     key = (objectid(data), sel)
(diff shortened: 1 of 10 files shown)


fhagemann:
I also want to mention that most of this code and, therefore, full  credit should go to @bene73 ! 🎉

src/legend_data.jl
```diff
@@ -299,10 +299,25 @@ function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol
             hvcard::Int = get(chmap[k].voltage.card, :id, -1)
```
theHenks, line 299: Let's move 
``` julia
cc4::StaticString{8} = get(chmap[k].electronics.cc4, :id, "")
            cc4ch::Int = get(chmap[k].electronics.cc4, :channel, -1)
            daqcrate::Int = get(chmap[k].daq, :crate, -1)
            daqcard::Int = chmap[k].daq.card.id
            hvcard::Int = get(chmap[k].voltage.card, :id, -1)
```
also into the detailed section.

theHenks: Also let's move
``` julia
enrichment::Unitful.Quantity{<:Measurement{<:Float64}} = if haskey(diodmap, k) && haskey(diodmap[k].production, :enrichment) measurement(diodmap[k].production.enrichment.val, diodmap[k].production.enrichment.unc) else measurement(Float64(NaN), Float64(NaN)) end *100u"percent"
            mass::Unitful.Mass{<:Float64} = if haskey(diodmap, k) && haskey(diodmap[k].production, :mass_in_g) diodmap[k].production.mass_in_g else Float64(NaN) end *1e-3*u"kg"
```
into that section

theHenks (changes requested):
Let's move the mentioned sections also into the `detailed` kwarg. Apart from that I am very happy with that!

theHenks:
I added my review: Let's move a couple more columns into the `detailed` section. Rest ist perfect!

fhagemann:
Should I put this already in this PR into `detailed` or should we make two PRs: this one (non-breaking) and the other one (breaking). With `detailed = false`, you would get the same output right now.

theHenks:
Same PR is fine. And the let's merge. Default should be `detailed::Bool=false`
I don't see this as a breaking change, since it is `dev` and not released. So we can still make a breaking release out of this.

fhagemann:
I moved the requested columns into the `detailed` section and also renamed `detailed` to `extended` (I feel like that name suits better, I also intuitively used it in the tests).

oschulz:
Can we make a similar calculation via SSD? Then we could compare in the package tests.

fhagemann:
> Can we make a similar calculation via SSD? Then we could compare in the package tests.

I could check, but only on a very fine grid.
There is active volume calculation with the SSD PointTypes.

oschulz:
> I could check, but only on a very fine grid.

It wouldn't need to run during package tests since there we can compare with a fixed reference value. But we could store the code to calculate/cross-check the reference value under `tests` for future re-calculation.

fhagemann:
I was able to compare the values using `get_active_volume` from this PR and SSD and I get the following
| Detector | `get_active_volume` by @bene73 | `get_active_volume` from SSD |
| -------- | ------- | ------- |
| V99000A |  `240.73197f0 cm^3` | `239.62573f0 cm^3` |
| B99000A | `200.78547f0 cm^3` | `199.84828f0 cm^3` |

Running the tests using SSD does not seem to take that long, so I can easily add this to the tests.

test/test_ext_ssd.jl
```diff
@@ -2,12 +2,91 @@
 
 using LegendDataManagement
 using Test
-
-using SolidStateDetectors:SolidStateDetector
+using SolidStateDetectors
 
 include("testing_utils.jl")
 
+function SolidStateDetectors.Simulation(l200, detname; medium::Union{String, Symbol} = "vacuum")
```
fhagemann, line 9: This might at some point become a function of the `SolidStateDetectors` extension of `LegendDataManagement`

PR #66, opened 2024-09-18
Implement active volume calculation for cracks

apearsonn:
Currently, the active volume calculation for detectors with cracks was not implemented.
I did the calculation and added the formulas for vertical and inclined cracks to the code.

The only thing missing is the handling of `fccd`, 
@fhagemann: could you add this?


fhagemann:
None of the test detectors have cracks, that's why the codecoverage fails.

theHenks:
Great job and thank you @apearsonn . Looks good to me and can be merged!

PR #67, opened 2024-09-18
Refactoring SolidStateDetectors.jl extension

bene73:
In this new implementation of the _SolidStateDetectors.jl_ extension, a config dictionary based on the detector metadata is created first and is afterwards used to create a `SolidStateDetectors.SolidStateDetector` or a `SolidStateDetectors.Simulation` object.

Previously it was only possible to create a `SolidStateDetectors.SolidStateDetector` object directly, which had no config dictionary. Thus, it was not possible to correctly save simulations in the HDF5 file format, which is now possible.

Except for the config dictionary, the resulting objects are the same as before.

---
The usage of the extension is the same as before, however now it is also possible to create a `SolidStateDetectors.Simulation` object:
```Julia
using LegendDataManagement, SolidStateDetectors
l200 = LegendData(:l200)
sim = SolidStateDetectors.Simulation(l200, :V00050A)
det = SolidStateDetectors.SolidStateDetector(l200, :V00050A)

```


test/test_ext_ssd.jl
```diff
@@ -2,91 +2,37 @@
...
             @test isapprox(active_volume_ssd, active_volume_ldm, rtol = 0.01)
+
+            # The creation via config files allows to save Simulations to files using LegendHDF5IO
+            lh5name = "$(detname).lh5"
+            isfile(lh5name) && rm(lh5name)
+            @test_nowarn ssd_write(lh5name, sim)
+            @test isfile(lh5name)
+            @test sim == ssd_read(lh5name, Simulation)
+            @test_nowarn rm(lh5name)
+            @test !isfile(lh5name)
```
fhagemann, line 35: These tests would previously fail:
Now that `Simulation` has a `config_dict`, the `sim` can actually be written and read using `LegendHDF5IO`.

fhagemann (approved):
Should be good to go now.

theHenks:
Great work, thanks @bene73 and @fhagemann . Could you maybe add the details about the `Simulation` object in the documentation for the extension. That would be awesome!

PR #71, opened 2024-09-25
Bump compat for SolidStateDetectors in [weakdeps] to 0.10.2

fhagemann:
This version bump is needed for the new implementation in the `SolidStateDetectors` extension to work.
(The charge drift model will otherwise not be parsed correctly)

Note: Wait with merging until `SolidStateDetectors` v0.10.1 is released.


fhagemann:
Re-run the tests tomorrow, when the `General` registry might be updated to include `SolidStateDetectors` v0.10.1

fhagemann:
Update: there was some serious type piracy in SolidStateDetectors, which should be fixed in 0.10.2 once it's released.

PR #72, opened 2024-11-01
Added the Radford Impurity Density model to the SSD extension

bene73:
Until now, each detector was characterised by a constant impurity density. By using the Radford Impurity Density model, we can get a more realistic parametrisation of the detectors.

The `impurity_density` field now contains:
- `name` :  "radford"
- `params` :  The five resulting parameters from the fit, as well as the detector offset

This implementation introduces a new package dependency to LegendDataManagement, namely LsqFit.


Project.toml
```diff
@@ -11,6 +11,7 @@ IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LsqFit = "2fda8390-95c7-5789-9bda-21331edee243"
```
fhagemann, line 14: @oschulz would you be ok with adding LsqFit as dependency until the fitted impurity density values make it into the metadata?

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -527,11 +554,27 @@ function create_SSD_config_dict_from_LEGEND_metadata(meta::PropDict, xtal_meta::
...
+        val = xtal_meta.impurity_measurements.value_in_1e9e_cm3                 # units: e/m^-3
+        fit_result = curve_fit(fit_model, pos, val, ones(Float64,5))
+
+        config_dict["detectors"][1]["semiconductor"]["impurity_density"] =  dicttype(
+            "name" => "radford", 
+            "parameters" => vcat(fit_result.param..., xtal_meta.slices[Symbol(meta.name[end])].detector_offset_in_mm / 1000)
+        )
+    else
+        @warn "No crystal metadata found for detector $(meta.name)"
+        # TODO: Implement a default impurity density for cases without crystal metadata
```
fhagemann, line 576: This would still need to be tested and done, I can take care of this

fhagemann:
We also might wanna add a keyword argument such that the user can choose whether they want to use the RadfordImpurityDensity from the crystal metadata or not

fhagemann:
I closed this by accident when trying to resolve the merge conflicts. Find the follow-up PR here: #74

PR #74, opened 2024-11-07
Add `RadfordImpurityDensity` model to the SSD extension

fhagemann:
Follow-up of #72 (got closed by accident when resolving the merge conflicts)


fhagemann:
This PR does, by default, **not** read the crystal metadata for the impurity density.
However, the default value is increased from `-1e9cm^-3` to `-5e9cm^-3`.

fhagemann:
Last changes (but should be good to go now):
- Changed the default impurity density to `-1e10cm^-3`
- Apply the correct order of magnitude (and sign for p-type) when reading the impurity density data
- Add tests for the impurity density read-in

hervasa2:
Note that I've already added this to legend-exp/LegendDetectorDesign.jl#1 (LDD) as `SegregationImpurityDensity`. Since LDD depends on LegendDataManagent where should this impurity density live? I think a better place for users to simulate detectors from metadata would be in LDD. Or maybe we should move the SSD extension there entirely and drop the LegendDataManagent dep?

oschulz:
I think we should move the impurity density fit to LDD. But the `RadfordImpurityDensity` struct and implementation must live in LegendDataManagement, otherwise it can't be used without depending on LDD. LDD should not become the primary package for simulating  pulses in LEGEND, that's what LegendGeSim is for.

fhagemann:
We could also move RadfordImpurityDensityModel to SolidStateDetectors.jl

hervasa2:
I would argue not to move RadfordImpurityDensityModel into SSD because we might want to keep it private. In general there will be a lot of overlap between LegendGeSim and LDD, so we should be careful on where to put all their shared functionality. One should not depend on the other.

oschulz:
> We could also move RadfordImpurityDensityModel to SolidStateDetectors.jl

Oh, yes, that would be best!

oschulz:
> because we might want to keep it private.

I don't think so - it would be open-source in any case, and we'd need to publish it as part of analysis papers anyway. We can ask David, of course, but I don't think he'll object.

hervasa2:
Added to SSD. See [JuliaPhysics/SolidStateDetectors.jl#473](https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/473)

hervasa2:
We should close this.

fhagemann:
Superseeded by #111

PR #78, opened 2024-11-26
More `ljl_propfunc` functionality

theHenks:
Increase possible functions in the allowed func set for the `ljl_propfunc`
In particular needed for the SiPM processing and the definition of the `LAr` anti-coincidence conditions in metadata.

### DIFF
--- a/src/ljl_expressions.jl
+++ b/src/ljl_expressions.jl
@@ -30,30 +30,33 @@ end
 export parse_ljlexpr
 
 
-const ljl_expr_allowed_heads = [:., :ref, :call, :macrocall, :||, :&&, :comparison, :if]
+const ljl_expr_allowed_heads = Symbol[:., :ref, :call, :macrocall, :||, :&&, :comparison, :if]
 
-const ljl_expr_allowed_funcs = Set([
+const ljl_expr_allowed_funcs = Set{Symbol}([
     :!,
     :(==), :<, :>, :>=, :<=, :!=,
-    :isapprox, :≈, :≈,
+    :isapprox, :≈,
     :in, :∈, :..,
-    :+, :-, :*, :/,
+    :+, :-, :*, :/, :div, :rem, :mod,
+    :|, :&, :xor,
     :^, :sqrt,
     :one, :zero, :identity,
     :abs, :abs2, :normalize, :norm,
     :exp, :exp2, :exp10, :log, :log2, :log10,
     :sin, :cos, :tan, :asin, :acos, :atan,
+    :min, :max,
     :isnan, :isinf, :isfinite,
     :all, :any, :broadcast,
-    :get, :getproperty,
-    :value, :uncertainty, :stdscore, :weightedmean,
-    :±, 
+    :sum, :prod, :minimum, :maximum, :mean,
+    :get, :getproperty, :getindex, :first, :last,
+    :haskey, :isempty, :length, :size,
     :(:), :Symbol, :String, :Int, :Float64, :Bool,
     :string, :parse,
+    :value, :uncertainty, :stdscore, :weightedmean, :±, 
     :DetectorId, :ChannelId
 ])
 
-const _ljlexpr_units = IdDict([
+const _ljlexpr_units = IdDict{Symbol,Expr}([
     :s => :(u"s"),
     :ms => :(u"ms"),
     :μs => :(u"μs"),


src/ljl_expressions.jl
```diff
@@ -44,11 +44,13 @@ const ljl_expr_allowed_funcs = Set([
     :sin, :cos, :tan, :asin, :acos, :atan,
     :isnan, :isinf, :isfinite,
     :all, :any, :broadcast,
-    :get, :getproperty,
+    :get, :getproperty, :getindex, :keys, :values, :pairs, :haskey,
+    :isempty, :length, :size, :first, :last, :findfirst, :findnext, :findall, :filter, :map, :reduce, 
+    :vcat, :hcat, :hvcat, :cat, :reshape, :transpose, :permutedims, :sort, :sortperm, :sortperm,
```
fhagemann, line 49: I see sortperm twice here

fhagemann (approved):
One duplicate, otherwise this looks fine to me

src/ljl_expressions.jl
theHenks: I removed the duplicate

theHenks:
@fhagemann I resolved the duplicates, let's merge.

oschulz:
@theHenks do you have some examples on how these will be used? I'm worried that we may be pushing this a bit too far, we shouldn't store algorithms in metadata strings ...

theHenks:
> @theHenks do you have some examples on how these will be used? I'm worried that we may be pushing this a bit too far, we shouldn't store algorithms (`:findnext, :findall, :filter, ...`) in metadata strings ...

I added these since I want to have variety in the description of the LAr anti-coincidence classifier and to express this in metadata. So far, we do, for example something like this hard-coded in `LegendDataManagement`:
``` julia
let a = a, m = m
        @pf (
            trig_pe = $trig_max .* m .+ a,
            trig_is_dc = [any(abs.($trig_pos_DC .- pos) .< 100u"ns") for pos in $trig_pos],
        )
 end
```
Since the discharge `DC` removal might require a more difficult software trigger layout, I wanna be flexible in the description of such cut (and have the option to set different ones in the `legend-metadata` e.g.
 ``` julia
  trig_is_dc = [any(abs.($trig_pos_DC[findall.($trig_max_DC .> trig_max_DC_threshold)] .- pos) .< 100u"ns") for pos in $trig_pos],
``` 
Etc.
The `reduce`, `vcat` and `hcat` functions are required since typically all `SiPM` analysis outputs are `VectorsOfVectors` and you need to flatten them after some initial cuts.
I am open to any other suggestions, @oschulz , but I want to avoid changing a LAr cut or PE reconstruction by opening a PR here every time we come up with a different scheme. Also, this will be even harder in my opinion if we start having different schemes for different `ChannelId`s.

oschulz:
We should not turn metadata into a generic programming language, that's a rabbithole. Values and simple value combinations ("calibration functions") yes (like the `100u"ns"`), but not algoritms, even small ones. We can have several algorithms defined in the code and let metadata select which one to apply, but putting algorithmic constructs into metadata with quickly become unmaintainable and also effectively means a huge (not so well) hidden code duplication.

theHenks:
I see your point. But then do you propose to put e.g. all the LAr cut generation into separate functions? 
I just find this hard to develop and iterate on because each change will require a PR and massively slow down the development. In addition, how do you want to handle different functions for different channels? Each channel then gets its own function?

fhagemann:
How about this: in order to meet the unblinding deadline, we merge this PR as is.
@theHenks provides examples of how it is used and, in the meantime, @oschulz can develop an alternative implementation that will at some point become the successor of this and make us revert this?

fhagemann:
I force-pushed to resolve the merge conflict introduced after merging #79

theHenks:
Yes, we will probably have to handle channels independently because of the cross-talk between SiPMs and HPGe channel. 

Can you tell us which functions you don't want to have in here? If I can't put this in here than I have to move the whole logic into `LegendEventAnalysis`. And have a lot of subcases.

oschulz:
> Can you tell us which functions you don't want to have in here?

Definitely not `:findfirst, :findnext, :findall, :filter, :map, :reduce, :vcat, :hcat, :hvcat, :cat, :reshape, :transpose, :permutedims, :sort, :sortperm` .

I'm also very doubtful about `:getproperty, :getindex, :keys, :values, :pairs, :haskey` .

oschulz:
As far as `:getindex` is concerned, sorry, I forgot that we didn't have support for `a[i]` yet, I'll add that.

oschulz:
In general I'd like  ljl-expressions to be constant-time (no searches, iterations, etc.).

oschulz:
Some of the requested functions (not `:keys, :values, :pairs, :findfirst, :findnext, :findall, :filter, :map, :reduce, :vcat, :hcat, :hvcat, :cat, :reshape, :transpose, :permutedims, :sort, :sortperm`) have been added in #83 now.

theHenks:
I am sorry, I messed up my git. Will fix that!

PR #79, opened 2024-12-03
Lookup table to generate `FileKey` Database

theHenks:
This PR includes look-up functionality to generate a `FileKey` database. The database can be generated with:
``` julia
l200 = LegendData(:l200)
ds = search_disk(DataSet, l200)
```
To search for a `FileKey` a LEGEND timestamp belongs to, you can use the `find_filekey` function. 
For convenience, plot recipes have been added to plot a specific timestamp, either from `phy` or from `cal` data.


ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -12,6 +12,19 @@ const ChannelOrDetectorIdLike = Union{ChannelIdLike, DetectorIdLike}
...
+        true
+    elseif LegendDataManagement._can_convert_to(DetectorId, det)
+        true
+    elseif LegendDataManagement._can_convert_to(DataTier, det)
+        true
+    else
+        @warn "Skipped $det since it is neither a valid `ChannelId`, `DetectorId` nor a `DataTier`"
+        false
+    end
+end
```
fhagemann, line 26: Is there any reason to pass `rsel` to this function? It does not seem to be used in the function.

fhagemann: This could also become something like
```julia
if LegendDataManagement._can_convert_to(ChannelId, det) ||
    LegendDataManagement._can_convert_to(DetectorId, det) ||
    LegendDataManagement._can_convert_to(DataTier, det)
    true
else
    @warn "Skipped $det since it is neither a valid `ChannelId`, `DetectorId` nor a `DataTier`"
    false
end
```

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -143,6 +156,7 @@ function LegendDataManagement.read_ldata(f::Base.Callable, data::LegendData, rse
     ch_keys = _lh5_data_open(data, rsel[1], rsel[2], "") do h
         keys(h)
     end
+    filter!(x -> _is_valid_channel_or_tier(data, rsel[2], x), ch_keys)
```
fhagemann, line 159: See comment above: `rsel[2]` does not seem to be used in `_is_valid_channel_or_tier`

ext/LegendDataManagementPlotsExt.jl
```diff
@@ -8,8 +8,10 @@ using PropDicts
 using Statistics
 using TypedTables
 using Unitful
+using Dates
 using Format
 using Measurements: value, uncertainty, weightedmean
+using LegendDataManagement
```
fhagemann, line 14: Would it be a lot of work to explicitly call only what is needed (`using LegendDataManagement: LegendData, FileKey, ChannelIdLike,...`).
This might make it easier to see if changes in `LegendDataManagement` might affect the plots, because they are not tested when the code is changed.

ext/LegendDataManagementPlotsExt.jl
```diff
@@ -131,7 +133,131 @@ using Measurements: value, uncertainty, weightedmean
...
+    end
+    plot_title := "$(fk.setup)-$(fk.period)-$(fk.run)-$(fk.category)"
+    for (p, p_wvf) in enumerate(plot_waveform)
+        @series begin
+            if p == 1
+                label := "$(channelinfo(data, fk, ch).detector) ($(ch))"
+            else
+                label := :none
+            end
+            colour := 1
```
fhagemann, line 163: I think its safer to use `color` here because `colour` is just an alias.

ext/LegendDataManagementPlotsExt.jl
```diff
@@ -131,7 +133,131 @@ using Measurements: value, uncertainty, weightedmean
...
+            for (c, chinfo_ch) in enumerate(chinfo)
+                @debug "Load $(chinfo_ch.detector)"
+                for (p, p_wvf) in enumerate(system[sys])
+                    @series begin
+                        if p == 1
+                            label := "$(chinfo_ch.detector) ($(chinfo_ch.channel))"
+                        else
+                            label := ""
+                        end
+                        colour := c
```
fhagemann, line 244: `colour`.

src/legend_data.jl
```diff
@@ -224,28 +232,78 @@ search_disk(DataCategory, l200.tier[:raw])
...
-    channelinfo(data::LegendData, sel::RunCategorySelLike; system::Symbol = :all, only_processable::Bool = false)
+    channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol = :all, only_processable::Bool = false, only_usability::Symbol = :all, extended::Bool = false)
+    channelinfo(data::LegendData, sel::RunCategorySelLike; system::Symbol = :all, only_processable::Bool = false, only_usability::Symbol = :all, extended::Bool = false)
 
 Get all channel information for the given [`LegendData`](@ref) and
 [`ValiditySelection`](@ref).
 """
-function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol = :all, only_processable::Bool = false, extended::Bool = false, verbose::Bool = true)
+function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol = :all, only_processable::Bool = false, only_usability::Symbol = :all, sort_by::Symbol=:detector, extended::Bool = false, verbose::Bool = true)
     key = (objectid(data), sel)
```
fhagemann, line 307: I would call this keyword `usability` rather than `only_usability`.

src/legend_data.jl
```diff
@@ -334,12 +392,22 @@ function channelinfo(data::LegendData, sel::AnyValiditySelection; system::Symbol
...
+    # apply filters and masks
     if !(system == :all)
         chinfo = chinfo |> filterby(@pf $system .== system)
     end
     if only_processable
         chinfo = chinfo |> filterby(@pf $processable .== true)
     end
+    if !(only_usability == :all)
+        chinfo = chinfo |> filterby(@pf $usability .== only_usability)
+    end
```
fhagemann, line 404: Because the way `only_processable` and `only_usability` work are somewhat different (`only_processable` will make you filter the `chinfo`, whereas `usability` will let you choose a specific subset)

src/utils/pars_utils.jl
```diff
@@ -38,29 +38,35 @@ Write validity for a given filekey.
...
+        elseif length(is_validity) == 1
+            @info "Merge old $pars_validTimeStamp $(string(category)) validity entry"
+            apply = unique(append!(Vector{String}(JSON.parse(validity_lines[first(is_validity)])["apply"]), apply))
+            validity_lines[first(is_validity)] = "{\"valid_from\":\"$pars_validTimeStamp\", \"category\":\"$(string(category))\", \"apply\":[\"$(join(sort(apply), "\", \""))\"]}"
+        end
+        # write validity
+        open(validity_filename, "w") do io
+            for line in sort(validity_lines)
+                println(io, line)
+            end
```
fhagemann, line 69: I am not an expert to judge what happens here, so I cannot comment on this.

src/ljl_expressions.jl
```diff
@@ -48,6 +48,7 @@ const ljl_expr_allowed_funcs = Set([
     :value, :uncertainty, :stdscore, :weightedmean,
     :±, 
     :(:), :Symbol, :String, :Int, :Float64, :Bool,
+    :string, :parse,
```
fhagemann, line 51: Where are you using `parse`?

Project.toml
```diff
@@ -55,6 +56,7 @@ LinearAlgebra = "<0.0.1, 1"
 MIMEs = "0.1, 1"
 Markdown = "<0.0.1, 1"
 Measurements = "2.2.1"
+OhMyThreads = "0.7"
```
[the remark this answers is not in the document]
theHenks: This package has basically 0 dependencies, so I vote in that case for the latest version to have the optimal speed. Are you fine with that too?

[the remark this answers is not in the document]
theHenks: Changed.

docs/src/extensions.md
```diff
@@ -1,5 +1,65 @@
...
+``` julia
+using Unitful, Dates, LegendDataManagement, Plots
+
+l200 = LegendData(:l200)
+
+ts = 1.6785791257987175e9u"s"
+
+ch = ChannelId(1104000)
+
+plot(l200, ts, ch)
```
fhagemann, line 45: why does this example need `Dates`?

docs/src/extensions.md
```diff
@@ -1,5 +1,65 @@
...
+```
+
+- `plot_tier`: The data tier to be plotted. Default is `DataTier(:raw)`.
+- `plot_waveform`: All waveforms to be plotted from the data. Default is `[:waveform_presummed]` which plots the presummed waveform.
+- `show_unixtime`: If set to `true`, use unix time instead of the datetime in the title. Default is `false`.
+
+If the channel is not give, the recipe automtically searches for the correct event in the data.
```
fhagemann, line 52: Typo: "given" is missing an n

src/utils/pars_utils.jl
theHenks: This is a bug fix which was done to prevent multiple workers to write the same validity file at the same time. I should have moved this to a separate PR, sorry.

docs/src/extensions.md
theHenks: It doesn't I forgot to remove

src/legend_data.jl
theHenks: I want to specifically mark that the user selects here only this `usability` and looses all others, so I want to leave that.

src/legend_data.jl
theHenks: s. other comment

fhagemann:
The introduction of `OhMyGod@>=0.5` requires `StructArrays@>=0.6`

ext/LegendDataManagementLegendHDF5IOExt.jl
theHenks: Done.
It might be required in the future to make the check more difficult since the upcoming data taking will change the channel map and therefore the `ChannelId`

ext/LegendDataManagementLegendHDF5IOExt.jl
theHenks: s. above

ext/LegendDataManagementPlotsExt.jl
theHenks: Let's do this as an issue. I think the whole plot recipe strategy we have right now needs a refactor

src/ljl_expressions.jl
theHenks: I forgot why. But same as for the others here. It extends the functionality for the future

fhagemann:
The last commit (setting `StructArrays@>=0.6.6`)  will fix the failing Downgrade test, because of a fix when materializing a `StructArray` in broadcast, see https://github.com/JuliaArrays/StructArrays.jl/pull/213.

theHenks:
Thanks, this looks good from my side nowy

fhagemann:
I cleaned up the commit history a bit (stashing the fixes into the initial commits and omitting the merge commits)

fhagemann:
@theHenks I wrote some tests, but I am abit confused by this commit message (cf74f7b):
image
because according to the tests, `partitioninfo` seems to return an `IdDict`.

PR #80, opened 2024-12-03
Add uncertainties to active volume calculation.

apearsonn:
I added code that reads the uncertainty of the fccd from the metadata to include with the active volume calculations. I took the correlated error (there is also uncorrelated error) which is up for discussion. The fccd values were also saved to the channelinfo table in units of mm.


oschulz:
> Do we want active_volume to ALWAYS be a Measurement? 

Do we always have an uncertainty for it, if we have a number?

fhagemann:
> > Do we want active_volume to ALWAYS be a Measurement?
> 
> Do we always have an uncertainty for it, if we have a number?

There is always a field uncertainty, but the value can be zero.
Just sometimes, there is no value for the FCCD at all in the metadata, but this is covered when reading in the FCCD.

fhagemann:
Closing this in favor of #87

PR #81, opened 2024-12-03
Include extras in SSD config file creation

apearsonn:
Adjust configuration of config dictionaries to include detectors that have extras (top groove, bottom cylinder, crack).


fhagemann:
I will have a look at this and see how it performs.
Let's also check if it can be merged without problems into #74.

hervasa2:
Yes it is. I can review it within the next couple months but if someone wants to step in sooned that would be great!

fhagemann:
Superseded by #157

PR #83, opened 2024-12-04
Support dot-based getproperty in ljl_propfunc

oschulz:
Also add support for square-brackets index access, the conditional ternary operator and functions `one`, `zero` and `identity`.


src/ljl_expressions.jl
```diff
@@ -95,7 +95,12 @@ _process_ljlexpr_impl(sym::Symbol, f_varsubst) = f_varsubst(sym)
 function _process_ljlexpr_impl(@nospecialize(expr::Expr), @nospecialize(f_varsubst))
     _process_inner = Base.Fix2(_process_ljlexpr_impl, f_varsubst)
     if expr.head in ljl_expr_allowed_heads
-        if expr.head == :call
+        if expr.head == :.
+            @info "foo"
```
fhagemann, line 99: This info seems like a leftover

PR #84, opened 2024-12-08
Bug fix `write_validity`

theHenks:
Fixed import of `Distributed` for validity creation.

### DIFF
--- a/src/utils/pars_utils.jl
+++ b/src/utils/pars_utils.jl
@@ -38,7 +38,7 @@ Write validity for a given filekey.
 function writevalidity end
 export writevalidity
 function writevalidity(props_db::LegendDataManagement.MaybePropsDB, filekey::FileKey, apply::Vector{String}; category::DataCategoryLike=:all)
-    remotecall_fetch(_writevalidity_impl, 1, props_db, filekey, apply; category=category)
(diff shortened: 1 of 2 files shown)


src/utils/pars_utils.jl
```diff
@@ -38,7 +38,7 @@ Write validity for a given filekey.
 function writevalidity end
 export writevalidity
 function writevalidity(props_db::LegendDataManagement.MaybePropsDB, filekey::FileKey, apply::Vector{String}; category::DataCategoryLike=:all)
-    remotecall_fetch(_writevalidity_impl, 1, props_db, filekey, apply; category=category)
```
fhagemann, line 41: I have no idea what this does.
If you extend the method in Distributed, you should add type annotations to the variables (one of which should be owned by LegendDataManagement to avoid type piracy)

theHenks: It is not an extension, I am just calling the function from `Distributed`. Since `LegendDataManagement` only imports `Distributed`, I did the same inside the `LDMUtils`

fhagemann: Sorry, the formatting on my phone confused me. Looks alright to me!

fhagemann (approved):

PR #85, opened 2024-12-10
Bug fix `lprops`

theHenks:
To properly parse units with e.g. special units from `UnitfulAtomic`, the function `unit_from_string` from `LegendDataTypes` is now used to avoid that.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -29,6 +29,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 TypedTables = "9d95f2ec-7b3d-5a63-8d20-e2491e220bb9"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
+UnitfulAtomic = "a7773ee8-282e-5fa2-be4e-bd808c38a91a"
 
 [weakdeps]
 LegendDataTypes = "99e09c13-5545-5ee2-bfa2-77f358fb75d8"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -29,16 +30,16 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 TypedTables = "9d95f2ec-7b3d-5a63-8d20-e2491e220bb9"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
+UnitfulAtomic = "a7773ee8-282e-5fa2-be4e-bd808c38a91a"
```
fhagemann, line 33: We do not need the direct `UnitfulAtomic` dependency anymore as it is hidden in `LegendDataTypes`.
I would remove it here.

fhagemann (approved):

oschulz:
Let's duplicate the `units_from_string` function in LegendDataManagement, instead of depending on LegendDataTypes.

This is a special situation: We want to spin off most of LegendHDFIO into a standalone registered package, and that function would move there then. So LegendDataManagement, which would not depend on that package, needs it's own version of the function anyway.

fhagemann:
> Let's duplicate the `units_from_string` function in LegendDataManagement, instead of depending on LegendDataTypes.
> 
> This is a special situation: We want to spin off most of LegendHDFIO into a standalone registered package, and that function would move there then. So LegendDataManagement, which would not depend on that package, needs it's own version of the function anyway.

This should be done now.

PR #86, opened 2024-12-13
Allow `partitioninfo` to filter categories

theHenks:
`partitioninfo` allows to filter for `analysis_runs` only to be able to parse `:phy` and `:cal`

### DIFF
--- a/src/utils/pars_utils.jl
+++ b/src/utils/pars_utils.jl
@@ -107,7 +107,7 @@ function get_partitionvalidity(data::LegendData, ch::ChannelIdLike, det::Detecto
     # unpack
     ch, det, part = ChannelId(ch), DetectorId(det), DataPartition(part)
     # get partition validity
-    partinfo = partitioninfo(data, ch, part; category=cal)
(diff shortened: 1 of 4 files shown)


src/utils/pars_utils.jl
```diff
@@ -107,7 +107,7 @@ function get_partitionvalidity(data::LegendData, ch::ChannelIdLike, det::Detecto
     # unpack
     ch, det, part = ChannelId(ch), DetectorId(det), DataPartition(part)
     # get partition validity
-    partinfo = partitioninfo(data, ch, part)
+    partinfo = partitioninfo(data, ch, part; category=cal)
```
fhagemann, line 110: I believe this should be `cat` instead of `cal`

fhagemann (commented):
Looks good to me, but I did not run the code. One minor comment

src/utils/pars_utils.jl
theHenks: Typo, sorry

PR #87, opened 2024-12-17
Add uncertainties for active volume calculation

fhagemann:
Alternative to #80 (which can be used once https://github.com/legend-exp/legend-detectors/pull/48 is merged and according dummy values are added to legend-testdata).

I am not sure if `Measurements` can deal with asymmetric errors, so right now I'm estimating the uncertainty as the maximum number quoted as `uncertainty` in the metadata (`pos` and `neg`).


fhagemann:
This implementation will only work if the latest metadata is loaded 
(with the FCCD values provided in https://github.com/legend-exp/legend-detectors/pull/48), so make sure to upgrade the detector metadata before using this

fhagemann:
@theHenks

PR #91, opened 2025-02-24
Add HPGeEnvironment

hervasa2:
Moved and modified Environment from LegendGeSim to LegendDataManagement. Now renamed to HPGeEnvironment.
Extended methods of `Simulation` and `SolidStateDetector` in the ssd extension accordingly.

```julia
filename = "V03422A.yaml"
env = HPGeEnvironment("LAr", 87u"K")
T = Float32
det = SolidStateDetector{T}(LegendData, 
    readlprops(filename), 
    LegendDataManagement.NoSuchPropsDBEntry("",[]),
    env
)
sim = Simulation{T}(LegendData, 
    readlprops(filename), 
    LegendDataManagement.NoSuchPropsDBEntry("",[]),
    env
)
```
Existing methods such as `sim = Simulation{T}(LegendData, filename)` call the new method with `env = HPGeEnvironment()` which sets the standard vacuum cryostat environment previously defined.


ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -86,12 +90,15 @@ function SolidStateDetectors.Simulation{T}(::Type{LegendData}, meta::AbstractDic
...
 function SolidStateDetectors.Simulation{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}) where {T<:AbstractFloat}
-    config_dict = create_SSD_config_dict_from_LEGEND_metadata(meta, xtal_meta)
-    return Simulation{T}(config_dict)
+    SolidStateDetectors.Simulation{T}(LegendData, meta, xtal_meta, HPGeEnvironment())
 end
 
+function SolidStateDetectors.Simulation{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}, env::HPGeEnvironment) where {T<:AbstractFloat}
+    config_dict = create_SSD_config_dict_from_LEGEND_metadata(meta, xtal_meta, env)
+    return Simulation{T}(config_dict)
+end
```
fhagemann, line 99: Can this not be combined to a single method, something like:
```julia
function SolidStateDetectors.Simulation{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}, env::HPGeEnvironment = HPGeEnvironment()) where {T<:AbstractFloat}
    SolidStateDetectors.Simulation{T}(LegendData, meta, xtal_meta, env)
end
```
?

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -46,7 +46,11 @@ function SolidStateDetectors.SolidStateDetector{T}(::Type{LegendData}, meta::Abs
...
 
 function SolidStateDetectors.SolidStateDetector{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}) where {T<:AbstractFloat}
-    config_dict = create_SSD_config_dict_from_LEGEND_metadata(meta, xtal_meta)
+    SolidStateDetectors.SolidStateDetector{T}(LegendData, meta, xtal_meta, HPGeEnvironment())
+end
+
+function SolidStateDetectors.SolidStateDetector{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}, env::HPGeEnvironment) where {T<:AbstractFloat}
+    config_dict = create_SSD_config_dict_from_LEGEND_metadata(meta, xtal_meta, env)
     return SolidStateDetector{T}(config_dict, SolidStateDetectors.construct_units(config_dict))
 end
```
fhagemann, line 55: Same thing here: 
```julia
function SolidStateDetectors.SolidStateDetector{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}, env::HPGeEnvironment = HPGeEnvironment()) where {T<:AbstractFloat}
    config_dict = create_SSD_config_dict_from_LEGEND_metadata(meta, xtal_meta, env)
    return SolidStateDetector{T}(config_dict, SolidStateDetectors.construct_units(config_dict))
end
```

src/hpge_environment.jl
```diff
@@ -0,0 +1,25 @@
...
+    struct HPGeEnvironment{T}
+
+Struct to describe the environment of a HPGe detector which are not set in metadata.
+
+## Parametric types
+* `T`: Temperature type
+
+## Fields
+* `medium::AbstractString`: Name of the medium.
+* `Temperature::T`: Temperature of detector/medium with units.
```
fhagemann, line 11: lower case `temperature`

fhagemann (commented):
I like this, see my minor comments.

theHenks:
The `read_lprops` function does not support `YAML` yet. Did you check this?

hervasa2:
@theHenks `YALM` compatibility was added in oschulz/PropDicts.jl#17

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -85,13 +85,12 @@ function SolidStateDetectors.Simulation{T}(::Type{LegendData}, meta::AbstractDic
...
-function SolidStateDetectors.Simulation{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}) where {T<:AbstractFloat}
-    config_dict = create_SSD_config_dict_from_LEGEND_metadata(meta, xtal_meta)
+function SolidStateDetectors.Simulation{T}(::Type{LegendData}, meta::PropDict, xtal_meta::Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}, env::HPGeEnvironment = HPGeEnvironment()) where {T<:AbstractFloat}
+    config_dict = create_SSD_config_dict_from_LEGEND_metadata(meta, xtal_meta, env)
     return Simulation{T}(config_dict)
 end
 
-
-function create_SSD_config_dict_from_LEGEND_metadata(meta::PropDict, xtal_meta::X; dicttype = Dict{String,Any}) where {X <: Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}}
+function create_SSD_config_dict_from_LEGEND_metadata(meta::PropDict, xtal_meta::X, env::HPGeEnvironment; dicttype = Dict{String,Any}) where {X <: Union{PropDict, LegendDataManagement.NoSuchPropsDBEntry}}
```
fhagemann, line 93: Maybe also give `env` the default `HPGeEnvironment` here.
This way, none of the changes made here are breaking and all existing code should still work.

src/hpge_environment.jl
```diff
@@ -0,0 +1,25 @@
...
+
+* `HPGeEnvironment()`.
+"""
+
+struct HPGeEnvironment{T<: Number}
+    medium::AbstractString
+    temperature::T
+    
+    HPGeEnvironment(m::AbstractString,t::T) where {T} = new{T}(m,t)
+end
```
fhagemann, line 23: If you explicitly call `ustrip` on `env.temperature`, you might wanna constrain the type `T` to 
`Unitful.Temperature{<:Number}` instead of just `Number`:
```julia
struct HPGeEnvironment{T<:Unitful.Temperature{<:Number}}
    medium::AbstractString
    temperature::T

    HPGeEnvironment(m::AbstractString,t::T) where {T} = new{T}(m,t)
end
```

fhagemann (commented):
Any reason why this is called `HPGeEnvironment`? In principle, I don't see why it has to have `HPGe` in its name..

fhagemann (approved):
Nice!

hervasa2:
We discussed with @oschulz. `Environment` would be too generic, specially to export.

fhagemann:
And something like `DetectorEnvironment`?

fhagemann:
I extended **all** methods in the SSD extension to pass `env` down to the core function and added tests to make sure that the environment is actually taken into account.

oschulz:
> Any reason why this is called HPGeEnvironment? In principle, I don't see why it has to have HPGe in its name..

I picked the name - I thought about making it non-HPGe specific, but we're not sure yet what field might be added later. We can always generalize it and rename with a typedef-compat without breaking API, but not the other way round.

theHenks:
> @theHenks `YALM` compatibility was added in oschulz/PropDicts.jl#17
> 
> 

I see nice! But then I think you should bump the `ProDicts` compat to `0.3`

fhagemann:
> > @theHenks `YALM` compatibility was added in [oschulz/PropDicts.jl#17](https://github.com/oschulz/PropDicts.jl/pull/17)
> 
> I see nice! But then I think you should bump the `ProDicts` compat to `0.3`

Not necessarily. This also works with JSON files. So, you just need to make sure to have the correct version of PropDicts depending on the file type you’re using.

theHenks:
> > > @theHenks `YALM` compatibility was added in [oschulz/PropDicts.jl#17](https://github.com/oschulz/PropDicts.jl/pull/17)
> > 
> > 
> > I see nice! But then I think you should bump the `ProDicts` compat to `0.3`
> 
> Not necessarily. This also works with JSON files. So, you just need to make sure to have the correct version of PropDicts depending on the file type you’re using.

I find this very confusing for users. I would rather want that the installation of a version of `LegendDataManagement` which supports `YAML` stuff, then also has the correct version of `PropDicts` for it. 
It happens quite often that they install `PropDicts` in their environment and don't update it so that a lower version is used. And then there will be big confusion why they can't read `YAML` files.

hervasa2:
Could we please merge? 🥺

fhagemann:
Do you also require a new release? Because the next release would be breaking because of other stuff happening in LegendDataManagement and I'm not sure whether we are ready for a new release now

hervasa2:
A release would be nice, but if its not ready I can live with it

oschulz:
This PR isn't breaking, but some other stuff already on main may be, so let's maybe wait with the release still.

PR #92, opened 2025-02-28
Updated `calibration_functions` and `evt_functions`

theHenks:
Updated calibration and evt functions to work properly with latest `metadata`, `pars` and `jlevt` level updates.

See [PR#13](https://github.com/legend-exp/LegendEventAnalysis.jl/pull/13) of `LegendEventAnalysis`

### DIFF
--- a/src/calibration_functions.jl
+++ b/src/calibration_functions.jl
@@ -7,19 +7,19 @@ _get_cal_values(pd::NoSuchPropsDBEntry, sel::AnyValiditySelection) = PropDicts.P
 ### HPGe calibration functions
 const _cached_get_ecal_props = LRU{Tuple{UInt, AnyValiditySelection}, Union{PropDict,PropDicts.MissingProperty}}(maxsize = 10^3)
 
-function _get_ecal_props(data::LegendData, sel::AnyValiditySelection)
+function _get_ecal_props(data::LegendData, sel::AnyValiditySelection; pars_type::Symbol=:rpars, pars_cat::Symbol=:ecal)
     key = (objectid(data), sel)
     get!(_cached_get_ecal_props, key) do
-        _get_cal_values(dataprod_parameters(data).rpars.ecal, sel)
+        _get_cal_values(dataprod_parameters(data)[pars_type][pars_cat], sel)
     end
 end
 
-function _get_ecal_props(data::LegendData, sel::AnyValiditySelection, detector::DetectorId)
-    get(_get_ecal_props(data, sel), Symbol(detector), PropDict())
+function _get_ecal_props(data::LegendData, sel::AnyValiditySelection, detector::DetectorId; kwargs...)
+    get(_get_ecal_props(data, sel; kwargs...), Symbol(detector), PropDict())
 end
 
-function _get_e_cal_propsfunc_str(data::LegendData, sel::AnyValiditySelection, detector::DetectorId, e_filter::Symbol)
-    ecal_props::String = get(get(get(_get_ecal_props(data, sel, detector), e_filter, PropDict()), :cal, PropDict()), :func, "e_max * NaN*keV")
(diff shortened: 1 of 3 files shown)


src/calibration_functions.jl
```diff
@@ -245,17 +247,50 @@ export get_ged_aoe_cut_propfunc
...
+function _dataprod_lar_cal(data::LegendData, sel::AnyValiditySelection, detector::DetectorId; pars_type::Symbol=:ppars)
+    @assert pars_type in (:ppars, :rpars) "pars_type must be either :ppars or :rpars"
+    dataprod_lar_config = _dataprod_spms_cal(data, sel).lar
+    merge(dataprod_lar_config[ifelse(pars_type == :ppars, :p_default, :default)], get(ifelse(pars_type == :ppars, get(dataprod_lar_config, :p, PropDict()), dataprod_lar_config), detector, PropDict()))
+end
+
 const _cached_get_larcal_props = LRU{Tuple{UInt, AnyValiditySelection}, Union{PropDict,PropDicts.MissingProperty}}(maxsize = 10^3)
 
-function _get_larcal_props(data::LegendData, sel::AnyValiditySelection)
+function _get_larcal_props(data::LegendData, sel::AnyValiditySelection; pars_type::Symbol=:ppars, pars_cat::Symbol=:sipmcal)
```
fhagemann, line 267: Just cross checking: `:sipmcal` is correct (looks like it was just `:sipm` before

fhagemann (commented):
Small syntax comment:
If you pass keyword arguments to functions and the variable you want to pass has the same name as the keyword argument, you can compactify your notation after the colon:

```julia
function_name(…; pars_type = pars_type)
```
Is equivalent to
```julia
function_name(…; pars_type)
```

src/calibration_functions.jl
theHenks: Yes. This was changed since before there was only one `pars` output for the sipms. So now we split into topical pars similar to the hpge with `ecal`

theHenks:
> Small syntax comment:
> 
> If you pass keyword arguments to functions and the variable you want to pass has the same name as the keyword argument, you can compactify your notation after the colon:
> 
> 
> 
> ```julia
> 
> function_name(…; pars_type = pars_type)
> 
> ```
> 
> Is equivalent to
> 
> ```julia
> 
> function_name(…; pars_type)
> 
> ```

Do you want me to change that? I haven't done it anywhere. So now at least it is consistent. I was also not aware of that notation.

src/ljl_expressions.jl
```diff
@@ -65,6 +65,7 @@ const _ljlexpr_units = IdDict{Symbol,Expr}([
     :MeV => :(u"MeV"),
     :keV => :(u"keV"),
     :eV => :(u"eV"),
+    :e => :(u"e_au"),
```
fhagemann, line 68: I believe that this change broke the [`aoe_optimization` processor](https://github.com/legend-exp/legend-julia-dataflow/blob/a51f772a61280cb2821d8057a3c85ab97cb95f80/processors/process_aoe_optimization.jl#L121-L123), because it uses `e` as the column name for energy in a `ljl_propfunc` and is wrongly interpreted..

With the newest version of `LegendDataManagement`, I get:
```julia
# qc_string = "(qc_label == 0.0 || qc_label == 7.0 || qc_label == 8.0 || qc_label == 10.0) && t50 > 45µs && t50 < 50µs && e > 1000.0 && e < 1000000.0 && isfinite(e)"
dep_sep_after_qc = (dep = dsp_dep[ljl_propfunc(qc_string).(dsp_dep)], 
                                sep = dsp_sep[ljl_propfunc(qc_string).(dsp_sep)]
                    )
```
```
┌ Error: Failed QC for DEP or SEP: MethodError(isless, (1000.0, e),
│ 0x0000000000006a6a)
└ @ Main In[8]:31

QC for DEP or SEP: MethodError(isless, (1000.0, e), 0x0000000000006a6a)

Stacktrace:
 [1] top-level scope
   @ In[8]:32
```

Going back to a version before this commit, this works without throwing an error.

PR #93, opened 2025-03-02
`LQ` calibration functions and `PSD` propfuncs

theHenks:
This PR includes the `LQ` calibration functions and `PSD` classifier defintions:
- `LQ` calibration and cut definition
- `psd_classifier` propfunc

**Attention**: This PR inverts the `A/E` and `LQ` cut logic: Any `classifier_cut` parameter is now `true` if the event should be cut. This makes the `PSD` part consisten with the rest of the `jlevt` level.

**Note**: This PR only works in combination with [PR#14](https://github.com/legend-exp/LegendEventAnalysis.jl/pull/14) in `LegendEventAnalysis`

### DIFF
--- a/src/calibration_functions.jl
+++ b/src/calibration_functions.jl
@@ -19,7 +19,7 @@ function _get_ecal_props(data::LegendData, sel::AnyValiditySelection, detector::
 end
 
 function _get_e_cal_propsfunc_str(data::LegendData, sel::AnyValiditySelection, detector::DetectorId, e_filter::Symbol; kwargs...)
-    ecal_props::String = get(get(get(_get_ecal_props(data, sel, detector; kwargs...), e_filter, PropDict()), :cal, PropDict()), :func, "e_max * NaN*keV")
+    ecal_props::String = get(get(get(_get_ecal_props(data, sel, detector; kwargs...), e_filter, PropDict()), :cal, PropDict()), :func, "$(e_filter) * NaN*keV")
     return ecal_props
 end
 
@@ -75,7 +75,7 @@ function _get_aoecal_props(data::LegendData, sel::AnyValiditySelection, detector
 end
 
 function _get_aoe_cal_propfunc_str(data::LegendData, sel::AnyValiditySelection, detector::DetectorId, aoe_type::Symbol; pars_type::Symbol=:ppars, pars_cat::Symbol=:aoe)
-    aoecal_props::String = get(_get_aoecal_props(data, sel, detector; pars_type=pars_type, pars_cat=pars_cat)[aoe_type], :func, "a_raw * NaN")
+    aoecal_props::String = get(_get_aoecal_props(data, sel, detector; pars_type=pars_type, pars_cat=pars_cat)[aoe_type], :func, "$(aoe_type) * NaN")
     return aoecal_props
 end
 
@@ -94,7 +94,7 @@ function _get_lqcal_props(data::LegendData, sel::AnyValiditySelection, detector:
 end
 
 function _get_lq_cal_propfunc_str(data::LegendData, sel::AnyValiditySelection, detector::DetectorId, lq_type::Symbol; pars_type::Symbol=:ppars, pars_cat::Symbol=:lq)
-    lqcal_props::String = get(_get_lqcal_props(data, sel, detector; pars_type=pars_type, pars_cat=pars_cat)[lq_type], :func, "lq * NaN")
+    lqcal_props::String = get(_get_lqcal_props(data, sel, detector; pars_type=pars_type, pars_cat=pars_cat)[lq_type], :func, "$(lq_type) * NaN")
     return lqcal_props
 end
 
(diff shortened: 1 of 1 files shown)


src/calibration_functions.jl
```diff
@@ -78,6 +79,25 @@ function _get_aoe_cal_propfunc_str(data::LegendData, sel::AnyValiditySelection,
...
+        _get_cal_values(dataprod_parameters(data)[pars_type][pars_cat], sel)
+    end
+end
+
+function _get_lqcal_props(data::LegendData, sel::AnyValiditySelection, detector::DetectorId; pars_type::Symbol=:ppars, pars_cat::Symbol=:lq)
+    get(_get_lqcal_props(data, sel; pars_type=pars_type, pars_cat=pars_cat), Symbol(detector), PropDict())
+end
+
+function _get_lq_cal_propfunc_str(data::LegendData, sel::AnyValiditySelection, detector::DetectorId, lq_type::Symbol; pars_type::Symbol=:ppars, pars_cat::Symbol=:lq)
+    lqcal_props::String = get(_get_lqcal_props(data, sel, detector; pars_type=pars_type, pars_cat=pars_cat)[lq_type], :func, "lq * NaN")
```
fhagemann, line 97: Is this related to issue https://github.com/legend-exp/LegendSpecFits.jl/issues/127, which results in NaN values and plotting problems?

src/calibration_functions.jl
```diff
@@ -102,12 +128,19 @@ detector.
 Note: Caches configuration/calibration data internally, use a fresh `data`
 object if on-disk configuration/calibration data may have changed.
 """
-function get_ged_psd_propfunc(data::LegendData, sel::AnyValiditySelection, detector::DetectorId; pars_type::Symbol=:ppars, pars_cat::Symbol=:aoe)
-    let aoe_types = collect(keys(_dataprod_aoe(data, sel, detector; pars_type=pars_type).aoe_funcs)), aoe_classifier = Symbol.(string.(keys(_dataprod_aoe(data, sel, detector; pars_type=pars_type).aoe_funcs)) .* "_classifier")
+function get_ged_psd_propfunc(data::LegendData, sel::AnyValiditySelection, detector::DetectorId; aoe_pars_type::Symbol=:ppars, aoe_pars_cat::Symbol=:aoe, lq_pars_type::Symbol=:ppars, lq_pars_cat::Symbol=:lq)
+    let aoe_types = Symbol.(_dataprod_aoe(data, sel, detector; pars_type=aoe_pars_type).aoe_types), aoe_classifier = Symbol.(_dataprod_aoe(data, sel, detector; pars_type=lq_pars_type).aoe_types .* "_classifier"), lq_types = Symbol.(_dataprod_lq(data, sel, detector; pars_type=lq_pars_type).lq_types), lq_classifier = Symbol.(_dataprod_lq(data, sel, detector; pars_type=lq_pars_type).lq_types .* "_classifier")
```
fhagemann, line 132:
```suggestion
    let aoe_types = Symbol.(_dataprod_aoe(data, sel, detector; pars_type=aoe_pars_type).aoe_types), aoe_classifier = Symbol.(_dataprod_aoe(data, sel, detector; pars_type=aoe_pars_type).aoe_types .* "_classifier"), lq_types = Symbol.(_dataprod_lq(data, sel, detector; pars_type=lq_pars_type).lq_types), lq_classifier = Symbol.(_dataprod_lq(data, sel, detector; pars_type=lq_pars_type).lq_types .* "_classifier")
```
?

fhagemann (commented):
Are there cases in which some function with `lq` in its name does not use `parse_cat = :lq`. Because if not, you might wanna remove the keyword argument `pars_cat` from these functions and define them within the functions.

Some LQ functions seem like code duplication and could be combined with A/E functions, but this can be done later.

theHenks:
> Are there cases in which some function with `lq` in its name does not use `parse_cat = :lq`. Because if not, you might wanna remove the keyword argument `pars_cat` from these functions and define them within the functions.
> 
> 
> 
> Some LQ functions seem like code duplication and could be combined with A/E functions, but this can be done later.

If I understand the question correctly you are asking if the `kwarg` is always used. I checked all and yes, they are everywhere were I put them in.

src/calibration_functions.jl
theHenks: No. The issue in `LegendSpecFits` is related to actual `NaN` values within the DSP output or the `LQ` calibration function. The `NaN` values here are put in here in case the detector is e.g. set to AC and no LQ values are generated. Also, this allows to use the same `calibrate_ged_channel_data` function independent of which `jlpar` are already generated and valid since missing `jlpar` just result in `NaN` values in the output

theHenks: I changed this to use the actual column of the filter we are looking up the `ljl_propfunc` for to be more consistent.

fhagemann:
> If I understand the question correctly you are asking if the `kwarg` is always used. I checked all and yes, they are everywhere were I put them in.

The question was whether in a function with LQ in its name, the keyword argument is always LQ.
But this can be dealt with once we remove code duplication

PR #95, opened 2025-03-05
Bug fix `calibration_functions` default case + Update `evt_functions`

theHenks:
This PR updates the `calibration_functions` to use `blmean` as default column in the DSP output which always exists. Furthermore, it updates the `evt_functions` with `get_ged_evt_is_valid_hit_properties` to allow selectable `is_valid_hit` properties.

**Note**: Part of this PR work only in combination with [PR#15](https://github.com/legend-exp/LegendEventAnalysis.jl/pull/15) in `LegendEventAnalysis`


fhagemann (approved):
Looks good

PR #98, opened 2025-03-10
Update to new metadata format

### LINKED ISSUE #90: Refactor to incorporate `YAML` based metadata
Since the latest update of `legend-metadata` the format changed to `YAML` including updates to the metadata logic.

fhagemann:
This can be addressed once https://github.com/legend-exp/LegendTestData.jl/pull/13 is merged and released:
Migrating to the new LEGEND metadata format.

In the end, this PR should close #90.


src/dataprod_config.jl
```diff
@@ -188,53 +175,127 @@ Base.Broadcast.broadcasted(f::typeof(partitioninfo), data::LegendData, ch::Vecto
...
+    end
+    runs
+end
+@inline parse_runs(s::AbstractString) = parse_runs([s])
+
+
+const _cached_analysis_runs = LRU{Tuple{UInt, DataCategoryLike}, StructVector{@NamedTuple{period::DataPeriod, run::DataRun}}}(maxsize = 10)
+function _analysis_runs(data::LegendData, cat::DataCategoryLike)
+    Table(sort(get!(_cached_analysis_runs, (objectid(data), cat)) do
+        aruns::PropDict = get(data.metadata.datasets.runlists.valid, Symbol(cat), PropDict())
```
fhagemann, line 245: Here, I use `:valid` in `runlists` as default. This could also be changed to a keyword argument to load a specific list of runs (e.g. `neutrino`, `prl25`...)

fhagemann (commented):
Breaking changes:
- update `load_validity` to new logic: basically a copy of https://github.com/gipert/dbetto/blob/983cf3bc8ec2c0a289a8f607e853583cca871075/src/dbetto/catalog.py#L104-L153
- update `_get_partitions` to new logic: read from `datasets/cal_groupings.yaml`
- update `partitioninfo` to new logic: expected `DetectorId` rather than `ChannelId` now. The old input, however, is still supported.
- update `get_partition_combined_periods` to new logic (the default method now also expects a `Vector{DetectorIdLike}`, rather than a `Vector{ChannelIdLike}` to follow the new metadata format. The old input is, however, still supported.
- add new method `parse_runs` to parse the new run format `rXXX..rYYY` in some files
- update `analysis_runs` to new logic: reading from `datasets/runlists.yaml`. Right now, it always reads `valid` by default. This could be changed to a keyword argument.
- To check if something is an analysis run, ALWAYS check `datasets/runlists/valid` (instead of checking if a calibration run is part of a partition)
- Redefine `DataPartition`: instead of just taking an `Int`, a `DataPartition` now consists of a `no`, `set` and `cat` (e.g. `calgroup001a --> cat == :calgroup, no == 1, set == :a`).
  This still supports passing only an `Int`, but it will use `cat = :calgroup` and `set = :a` as default, i.e. `DataPartition(1) => calgroup001a`
- read `processable` and `usability` from `datasets/statuses` rather than `dataprod/config/analysis`
- changed default filetype from `JSON` (or `JSONL`) to `YAML`
- update tests to run on testdata with new metadata format

fhagemann:
What is still missing:
- [ ] `_writevalidity_impl` when writing validities, still needs to be able to write `mode`, according to the new format
- [ ] check if there is no category, or if it is `all`, that the data is handled the correct way
- [ ] check what happens if a `validity.yaml` has same `valid_from` but different files in the append
- [ ] `write_validity` for partitions should be updated to the new logic, to use `replace` rather than listing ALL files all the times (look at `ppars/aoe` for example)
- [ ] Sort entries of YAML validity files, either when creating them or when reading them.
- [ ] Allow `runinfo` to take a name (`valid`, `neutrino`, `prl25`)
- [ ] Check that `partitioninfo` has all fields (not just `cal`, `phy`, `fft`, but also `cos`, `lac`)
- [ ] Check if `get_partition_channelinfo` was updated and works
- [ ] Restrict `DataPartition` `cat` to be a `DataCategory`, that can be e.g. `:cal`, `:calgroup`, `:calpart`

In the actual dataflow:
- [ ] Update the validity logic of all validity files (requires some automatized function, because manually this is UNFEASIBLE)

theHenks (changes requested):

theHenks (changes requested):

theHenks (commented):
Why are there missing enrichments in the metadata? This should have been fixed I guess. At least in the latest `json` data productions there was always an enrichment value available.

src/legend_data.jl
jonasschlegel, line 1: For know this should be a workaround for the extended channelinfo.
Not all enrichments are available: 
name: V03421A
type: icpc
production:
  manufacturer: Mirion
  order: 3
  crystal: "421"
  slice: A
  enrichment:
    val: null
    unc: 0.005
  passivation: true
  reprocessing: false
  mass_in_g: 2700
  impcc:
    array:
      value_in_1e9e/cm3:
        - 0
        - 0
      dist_from_contact_in_mm:
        - 0
        - 0
    function_pars:
      z0_in_1e10e/cm3: 0
      gradient_in_1e10e/cm4: 0
      quadratic_in_1e10e/cm5: 0
  delivered: "2024-03-13"
  I think we can push this for now and I will open a pull request in the metadata soon.

fhagemann:
Please do not post confidential data here.

theHenks (changes requested):
Let's change this for the enrichment. I will additionally add changes to the `readlprops` function to avoid `trim_null` in the `YAML` cases.

theHenks (changes requested):
Perfect! I will go ahead and merge this as a new release 0.6.0 while keeping the old json version in a legacy branch

theHenks (approved):

theHenks (approved):

PR #107, opened 2025-03-22
Update active volume calculation

fhagemann:
There were some inconsistencies with how the fccd was handled.. this should be fixed now.

I will be attaching some notes from my iPad for reference.
[Active volume.pdf](https://github.com/user-attachments/files/19404904/Active.volume.pdf)


theHenks (approved):
Thanks @fhagemann , I will go ahead and merge

PR #109, opened 2025-03-24
Fix base method of `get_exposure`

fhagemann:
This should return correct values, especially for detectors in which the `usability` changed between runs or periods.


theHenks (approved):
Thanks!

PR #114, opened 2025-04-28
Parallel `read_ldata`

theHenks:
This PR allows the use of `read_ldata` to read multiple files in parallel. The function `read_ldata` is extended by two additional `kwargs`:
- `parallel`: read files in parallel, default `false`
- `wpool`: worker pool to use for parallel read, only applies if `parallel = true`, default `default_worker_pool()` containing all workers 
The function automatically parallels all runs, periods and files for maximum speed.

PR #116, opened 2025-05-05
Move LegendHDF5IO tests to `test_ext_legendhdf5io.jl`

fhagemann:
The tests on Julia 1 fail because after #114, `LegendHDF5IO` is loaded before `SolidStateDetectors`, which causes a deadlock in `SolidStateDetectors`. To avoid this, I have moved all tests that require `LegendHDF5IO` to the respective test file `test_ext_legendhdf5io.jl`.

There is also a PR onto SSD that will fix this deadlock concurrency (https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/481), but this will fix it here also for the Downgrade tests.


theHenks:
Thank you o much @fhagemann ❤️

PR #117, opened 2025-05-07
Patch `PropsDB` overrides

theHenks:
This PR allows the `PropsDB` to use properties from overrides paths even in the case when there are no props existing under th `_base_path`. This is in particluar useful for `jlpars` which are disitrbuted via the `jldataprod-overrides` repository and not generated within the `dataflow`


fhagemann (approved):
Such a beautiful PR, great work!

PR #119, opened 2025-05-12
Define hierarchy of metadata values for dead layer thickness and bias voltage

hervasa2:
All contacts should be given a thickness in SSD to prevent charge drift convergence errors. A hierarchy was chosen for which metadata values for dead layer thickness and bias voltage should be used.


ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -93,17 +93,16 @@ end
...
-
     gap = 1.0 # to ensure negative volumes do not match at surfaces
 
-    dl_thickness_in_mm = haskey(meta.characterization.manufacturer, :dl_thickness_in_mm) ? meta.characterization.manufacturer.dl_thickness_in_mm : 0
+    dl_thickness_in_mm = if hasproperty(meta.characterization, :combined_0vbb_analysis) && meta.characterization.combined_0vbb_analysis.fccd_in_mm.value > 0
+        meta.characterization.combined_0vbb_analysis.fccd_in_mm.value
+    elseif hasproperty(meta.characterization.manufacturer, :dl_thickness_in_mm) && meta.characterization.manufacturer.dl_thickness_in_mm > 0
+        meta.characterization.manufacturer.dl_thickness_in_mm
+    else
+        1.0 # default value
```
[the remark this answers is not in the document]
fhagemann: I also kind of like the idea to have a default value defined somewhere outside, so that it can be easily changed.

ext/LegendDataManagementSolidStateDetectorsExt.jl
fhagemann, line 1: The FCCD and voltage hierarchy implemented here should be documented somewhere in the documentation (or maybe even shown as `@info` output for verbose simulations), so that the user can have some idea to understand what has happened without having to look into the source code.

hervasa2: Should we add a verbose keyword for all constructors `SolidStateDetectors.SolidStateDetector(data::LegendData, detector::DetectorIdLike, env::HPGeEnvironment = HPGeEnvironment(); verbose = true)`
And then print out a nice info with all the values set?

fhagemann: I would welcome this. Do you want to add this in this PR or a follow-up one?

fhagemann, line 1: To avoid a lot of output (we read most of this in the `channelinfo` for the LEGEND analysis), I would vote for setting the `verbose` keyword argument to false.

hervasa2: Which constructor is used there? Maybe we could set the constructors we use internally to false and the user facing ones to true?

fhagemann: Can you still add a small section on the FCCD and voltage hierarchy somewhere in the `docs`?

fhagemann (commented):
Would this also require a bump in the SSD version in [compat], once the changes in https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/491 are merged and released?

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -634,17 +634,25 @@ function create_SSD_config_dict_from_LEGEND_metadata(meta::PropDict, xtal_meta::
     
     slice = Symbol(meta.name[end])
     config_dict["detectors"][1]["semiconductor"]["impurity_density"] = if hasproperty(xtal_meta,:impurity_curve) && hasproperty(xtal_meta.slices, slice)
+        impurity_scale =  hasproperty(xtal_meta.impurity_curve.corrections, :scale) ? xtal_meta.impurity_curve.corrections.scale : 1.0
+        impurity_offset = hasproperty(xtal_meta.impurity_curve.corrections, :offset) ? xtal_meta.impurity_curve.corrections.offset * -1e6 : 0.0 ## 1e9cm^-3 -> mm^-3
```
fhagemann, line 638: Why minus? `-1e6`. To have it ALWAYS positive, when talking about p-type detectors?

fhagemann: Can this be simplified to something like (test if it works):
```suggestion
        impurity_offset = get(xtal_meta.impurity_curve.corrections, :offset, 0) * -1e6
```

hervasa2: > Why minus? `-1e6`. To use positive values when talking about p-type detectors? What if one were to try to make it "more n-type"?

Because everything in the metadata is positive. Thus a positive value in `xtal_meta.impurity_curve.corrections.offset` will result in more p-type impurities, and a negative value will make it more n-type.

ext/LegendDataManagementSolidStateDetectorsExt.jl
hervasa2: What do we do about this?

hervasa2:
> Would this also require a bump in the SSD version in [compat], once the changes in [JuliaPhysics/SolidStateDetectors.jl#491](https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/491) are merged and released?

Yes, but it can still be used without the SSD release. Any impurity corrections from the dict passed to ssd would just be ignored.

fhagemann:
> > Would this also require a bump in the SSD version in [compat], once the changes in [JuliaPhysics/SolidStateDetectors.jl#491](https://github.com/JuliaPhysics/SolidStateDetectors.jl/pull/491) are merged and released?
> 
> Yes, but it can still be used without the SSD release. Any impurity corrections from the dict passed to ssd would just be ignored.

But then you might get inconsistent results.
I would prefer to have this versioned properly.

ext/LegendDataManagementSolidStateDetectorsExt.jl
fhagemann: We are actually not calling the explicit `SolidStateDetector` or `Simulation` constructor defined here, so it should not be an issue. I'm gonna resolve this.

fhagemann (approved):

PR #122, opened 2025-06-03
period channelinfo function

verenaaur:
- I added a 'period_channelinfo' function to the legend_data.jl file.
- This function takes a period as an argument instead of a filekey.
- It then creates a combined channelinfo of all runs in this period. It first merges all channelinfos, then creates arrays in each column for each detector (besides 'detector') with all entries from the different runs. These arrays are filtered in the next step: in columns with constant entries, only one entry remains. In columns with changing flags, e.g. "valid" or "present" in "low_aoe_status" the 'best' flag is chosen (implemented in the 'get_hierarchies' function).
- kwargs and filters ( "|> filterby...") can be used as usual.

### DIFF
--- a/src/legend_data.jl
+++ b/src/legend_data.jl
@@ -589,11 +589,11 @@ function get_hierarchies(extended::Bool)
         :usability => [:on, :ac, :off],
         :is_blinded => [true, false],
         :psd_usability => [:on, :off],
-        :low_aoe_status => [:valid, :present, :missing],
-        :high_aoe_status => [:valid, :present, :missing],
-        :lq_status => [:valid, :present, :missing],
-        :ann_status => [:valid, :present, :missing],
-        :coax_rt_status => [:valid, :present, :missing]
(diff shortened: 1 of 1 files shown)


theHenks (changes requested):
At a first look, this looks good. However, I suggest the following:

- Let's not introduce a new function named `period_channelinfo`. I highly recommend the dispatch on the existing function so that the argument is a `PeriodSelLike`.
- All helper functions should start with a `_` so that it is clear that they are not meant to be used outside of the module
- Try to add types and comments to all arguments and function. Type safety and stability is important when this function will later be called very often.
- A way to cache also this `channelinfo` with the help of the `LRUCache` should definitely be done
- I think a lot of the code can be simplified and reduced by using more `Tables` and `NamedTuple` operations. I will try to help and support here once I had a look into the code in more details

theHenks:
Closed in favour of #220

PR #127, opened 2025-07-03
Update LegendDataManagementLegendHDF5IOExt.jl

jonasschlegel:
This pull request improves the robustness and flexibility of the read_ldata function in LegendDataManagementLegendHDF5IOExt.jl to support both legacy and new LEGEND raw data formats. In recent data (e.g., from period 14 onwards), raw HDF5 files use detector names as top-level keys, whereas older data uses channel IDs. The previous implementation of read_ldata always attempted to map between channel IDs and detector names, which caused errors or unnecessary complexity when the key already matched the file structure.

The central read_ldata function now first checks if the provided key (channel ID or detector name) exists directly in the HDF5 file. Only if the key is not found, the function attempts to map: Channel ID → Detector Name
Detector Name → Channel ID
If neither mapping is successful, a clear error or warning is raised. The function is now fully backwards compatible and works transparently for both old and new data formats.

### DIFF
--- a/ext/LegendDataManagementLegendHDF5IOExt.jl
+++ b/ext/LegendDataManagementLegendHDF5IOExt.jl
@@ -120,75 +120,64 @@ _load_all_keys(arr::AbstractArray, n_evts::Int=-1) = arr[:][if (n_evts < 1 || n_
 _load_all_keys(t::Table, n_evts::Int=-1) = t[:][if (n_evts < 1 || n_evts > length(t)) 1:length(t) else rand(1:length(t), n_evts) end]
 _load_all_keys(x, n_evts::Int=-1) = x
 
-function LegendDataManagement.read_ldata(
-    f::Base.Callable,
-    data::LegendData,
-    rsel::Tuple{DataTierLike, FileKey, ChannelOrDetectorIdLike};
-    n_evts::Int = -1,
-    ignore_missing::Bool = false,
-    parallel::Bool = false,
-    wpool::WorkerPool = default_worker_pool()
-)
+# --- Helper function for robust Channel/Detector mapping ---
+function _find_valid_key(h, data::LegendData, filekey, ch_str::AbstractString)
+    # 1. Directly present as key?
+    if haskey(h, ch_str)
+        return ch_str
+    end
+    # 2. ChannelId → DetectorName
+    if LegendDataManagement._can_convert_to(ChannelId, ch_str)
+        chinfo = channelinfo(data, (filekey.period, filekey.run, filekey.category, ch_str))
+        detname = string(chinfo.detector)
+        if haskey(h, detname)
+            return detname
+        end
+    end
+    # 3. DetectorName → ChannelId
+    if LegendDataManagement._can_convert_to(DetectorId, ch_str)
+        chinfo = channelinfo(data, (filekey.period, filekey.run, filekey.category, ch_str))
+        chanid = string(chinfo.channel)
+        if haskey(h, chanid)
+            return chanid
+        end
+    end
+    return nothing
+end
+
+function LegendDataManagement.read_ldata(f::Base.Callable, data::LegendData, rsel::Tuple{DataTierLike, FileKey, ChannelOrDetectorIdLike}; n_evts::Int=-1, ignore_missing::Bool=false, parallel::Bool=false, wpool::WorkerPool=default_worker_pool())
     tier, filekey, ch = DataTier(rsel[1]), rsel[2], rsel[3]
     _lh5_data_open(data, tier, filekey, ch) do h
         ch_str = string(ch)
-        correct_key = nothing
-        # 1. Check if the key exists directly in the file (works for both channel IDs and detector names)
-        if haskey(h, ch_str)
-            correct_key = ch_str
-        else
-            mapped = false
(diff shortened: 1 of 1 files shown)


ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -120,35 +120,70 @@ _load_all_keys(arr::AbstractArray, n_evts::Int=-1) = arr[:][if (n_evts < 1 || n_
 _load_all_keys(t::Table, n_evts::Int=-1) = t[:][if (n_evts < 1 || n_evts > length(t)) 1:length(t) else rand(1:length(t), n_evts) end]
 _load_all_keys(x, n_evts::Int=-1) = x
 
-function LegendDataManagement.read_ldata(f::Base.Callable, data::LegendData, rsel::Tuple{DataTierLike, FileKey, ChannelOrDetectorIdLike}; n_evts::Int=-1, ignore_missing::Bool=false, parallel::Bool=false, wpool::WorkerPool=default_worker_pool())
-    tier, filekey, ch = DataTier(rsel[1]), rsel[2], if !isempty(string((rsel[3]))) _get_channelid(data, rsel[2], rsel[3]) else rsel[3] end
+function LegendDataManagement.read_ldata(
```
theHenks, line 123: Why is this line a change? This looks like just reformatting or did I miss something?

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -120,35 +120,70 @@ _load_all_keys(arr::AbstractArray, n_evts::Int=-1) = arr[:][if (n_evts < 1 || n_
...
+    tier, filekey, ch = DataTier(rsel[1]), rsel[2], rsel[3]
     _lh5_data_open(data, tier, filekey, ch) do h
-        if !isempty(string((ch))) && !haskey(h, "$ch")
-            if ignore_missing
-                @warn "Channel $ch not found in $(basename(string(h.data_store)))"
-                return nothing
-            else
-                throw(ArgumentError("Channel $ch not found in $(basename(string(h.data_store)))"))
+        ch_str = string(ch)
+        correct_key = nothing
```
theHenks, line 135: This is not type safe. This function will be called a lot within a single `read_ldata` call in particular if you load from a load of files, so type safety should be taken into account.

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -120,35 +120,70 @@ _load_all_keys(arr::AbstractArray, n_evts::Int=-1) = arr[:][if (n_evts < 1 || n_
...
+        # 1. Check if the key exists directly in the file (works for both channel IDs and detector names)
+        if haskey(h, ch_str)
+            correct_key = ch_str
+        else
+            mapped = false
+            # 2. Try mapping channel ID → detector name 
+            try
```
theHenks, line 142: There should be a way to do the mapping without a huge `try catch` closure in here. This will also impact the speed (s. benchmark statement)

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -120,35 +120,70 @@ _load_all_keys(arr::AbstractArray, n_evts::Int=-1) = arr[:][if (n_evts < 1 || n_
...
             end
         end
+        # Load the data 
         if f == identity
-            if !isempty(string((ch)))
-                _load_all_keys(h[ch, tier], n_evts)
-            else
-                _load_all_keys(h[tier], n_evts)
-            end
+            _load_all_keys(h[correct_key, tier], n_evts)
```
theHenks, line 182: I don't fully understand the code. What happens in the empty string case? This is e.g. necessary when you load `jlevt` files.

theHenks (changes requested):
I would highly appreciate type safety here. Maybe the whole channel mapping should be offloaded into a separate helper function?
Some general remarks:
- Let's try to avoid any `try catch` closures here
- Did you benchmark the function after the change? There is e.g. the `@benchmark` macro from `BenchMarkTools.jl` which should give you some insides into the performance of your code.
- Do we really need to do this `ChannelId` mapping every time we load an individual `FileKey`? I am imagining the file layout across multiple runs a user wants to load should be consistent within the files. So it should be enough to check this once and then use the same access for all files of a run. 
- What happens in the case there is a `DetectorId` key as well as a `ChannelId` key in the `lh5` file? As far as I remember, it was once discussed to keep the `ChannelId` access patterns as virtual datasets within each `lh5` file. Is this the case for the current files of the new period?

theHenks:
@jonasschlegel Thanks for the PR. I submitted my review. I still have some open changes regarding the `read_ldata` function in general to improve speed and performance as well as the stability. This PR will also allow more complicated filterby and `PropertyFunction` access patterns while loading more data. 
If you are ok with that, we could first merge the speed improvement and updates and then come back to this.

theHenks:
@jonasschlegel I think this PR needs a rebase after the updates on `main`. Right now, there are some conflicts.

theHenks:
@jonasschlegel Why was this closed? I don't like the idea of PRs being closed without any comment nor reason in the middle of a review. This still seems to be an issue, so I would reopen.

Issue #120, opened 2025-05-15
Avoid `readvalidity` reading from a validity file that is being written


fhagemann:
We need to respect the `ReentrantLock` (`_writevalidity_lock`) for `readvalidity`.
We should not read while something is writing

PR #129, opened 2025-07-16
Finish update to the new metadata format

DaGeibl:
This PR adds the open tasks that were missing in Felix YAML PR (https://github.com/legend-exp/LegendDataManagement.jl/pull/98). The fixes introduced are namely:

- Restricted the Datapartition cat to be a DataCategory
- runinfo now can take a dataset name e.g. valid, neutrino
- also updated the is analysis run logic to use groupings instead of Period
- Updated partitioninfo and channelinfo to work with groupings instead of Partition
- updated the writevalidity to produce yaml files in the new format
-  checked if there is no category, or if it is all, that the data is handled the correct way
      -> handles the correct way, all is default if invalid category is set. phy is treated as invalid category.
- checked what happens if a validity.yaml has same valid_from but different files in the append
      -> first mentioned file is apllied first---> intended behavior

MISSING:
What is missing is the Write/Read lock implementation fixing issue (https://github.com/legend-exp/LegendDataManagement.jl/issues/120). I noticed implementing this should be easy if one uses Julia version 1.11, as this version introduces the Base.ReentrantRWLock() function, with which this should be doable.
I wanted to ask if it is an issue to require Julia 1.11 to run the data management.

This PR is also still work in progress. I want to test the changes with the dataflow to see if something breaks.
Also, while I updated the logic of how the writevalidity writes validities to remove duplicates, I think it still can be improved using the modes — that’s something I also want to test with the partition processors.

I noticed some tests are failing since the testdata only has cal groupings, but no phy groupings.
The other failing tests i still Need to Check, i probably have not updated the function Arguments of the functions in the Test Files. I will fix them asap.


theHenks:
@DaGeibl Thanks for the PR. Do you wanna start the review already? This sounds to me like it is still in the draft stage. I suggest however to start with an early review already to see how far we get.

fhagemann:
I can review this, but I would do so only when the tests are passing, so everything works as expected.
So please also update the tests (and, if needed, `LegendTestData`, even though this should be fine and according to the new data already), which makes it also easier to judge what breaking changes this PR introduces.

src/dataprod_config.jl
```diff
@@ -279,83 +291,133 @@ function analysis_runs end
...
                 if haskey(ri, cat)
                     fk = ifelse(haskey(ri[cat], :start_key), FileKey(data.name, period, run, cat, Timestamp(get(ri[cat], :start_key, 1))), missing)
-                    is_ana_run::Bool = (; period, run) in _analysis_runs(data, cat) && !ismissing(fk)
+                    is_ana_run = if cat == :phy
+                        (; period, run) in phy_groupings && !ismissing(fk)
+                    elseif cat == :cal 
+                        (; period, run) in cal_groupings && !ismissing(fk)
+                    else
+                        true
+                    end
```
fhagemann, line 386: Does this only apply to `phy` and `cal`? If not, this could probably be combined to something like
```suggestion
                    is_ana_run::Bool = (; period, run) in _groupings_runs(data, cat) && !ismissing(fk)
```

src/dataprod_config.jl
```diff
@@ -279,83 +291,133 @@ function analysis_runs end
...
         end
-        periods_and_runs = [[make_row(p, r, ri) for (r, ri) in rs] for (p, rs) in rinfo]
-        flat_pr = sort(StructArray(vcat(periods_and_runs...)::Vector{@NamedTuple{period::DataPeriod, run::DataRun, cal::nttype, phy::nttype, fft::nttype}}))
-        Table(merge(columns(flat_pr), (cal = Table(StructArray(flat_pr.cal)), phy = Table(StructArray(flat_pr.phy)), fft = Table(StructArray(flat_pr.fft)))))
+        periods_and_runs = [[make_row(p, r, ri) for (r, ri) in rs if any((cat -> (cat, p, DataRun(r)) in allowed), (:cal, :phy, :fft, :cos, :lac))] for (p, rs) in rinfo]
+        flat_pr = sort(StructArray(vcat(filter(!isempty, periods_and_runs)...)::Vector{@NamedTuple{period::DataPeriod, run::DataRun, cal::nttype, phy::nttype, fft::nttype, cos::nttype, lac::nttype}}))
+        Table(merge(columns(flat_pr), (cal = Table(StructArray(flat_pr.cal)), phy = Table(StructArray(flat_pr.phy)), fft = Table(StructArray(flat_pr.fft)), cos = Table(StructArray(flat_pr.cos)), lac = Table(StructArray(flat_pr.lac)))))
```
fhagemann, line 396: Maybe we can define a list of available `DataCategory` somewhere and iterate through this here.
This might also make it easier to add new categories, once they arise.

src/dataprod_config.jl
```diff
@@ -279,83 +291,133 @@ function analysis_runs end
...
     runinfo(data::LegendData)::Table
     runinfo(data::LegendData, runsel::RunSelLike)::NamedTuple
     runinfo(data::LegendData, filekey::FileKey)::NamedTuple
 
-Get the run information for `data`.
+Get the run information for `data` based on various selection criteria.
+
+# Arguments
+- `data::LegendData`: The dataset to query run information from.
+- `runlist::Symbol = :valid`: Specifies which runlist from `data.metadata.datasets.runlists` to use for filtering. Defaults to `:valid`.
```
fhagemann, line 343: `runlist` seems to be a keyword argument and not an argument

```suggestion

# Keyword Arguments
- `runlist::Symbol = :valid`: Specifies which runlist from `data.metadata.datasets.runlists` to use for filtering. Defaults to `:valid`.
```

src/props_db.jl
```diff
@@ -272,7 +272,9 @@ function _load_validity(validity_path::AbstractString; mode_default::AbstractStr
     if isfile(validity_path)
         # TODO: Cannot use readlprops here because this can be a Vector of PropDicts and not a PropDict
         #       Can readlprops be updated so that we can drop the explicit YAML dependency here?
-        entries = PropDict.(YAML.load_file(validity_path))
+        raw = YAML.load_file(validity_path)
+        entries = PropDict.(raw === nothing ? [] : raw)
+
```
fhagemann, line 277: Look into this `TODO`
@oschulz

src/utils/pars_utils.jl
```diff
@@ -41,45 +41,87 @@ function writevalidity(props_db::LegendDataManagement.MaybePropsDB, filekey::Fil
...
+                    println(io, "  mode: ", e["mode"])
+                    println(io)  # newline between entries
+                end
             end
         end
+
+        # emit valid YAML, with each Dict’s keys in insertion order
+        write_entries_ordered(dst, entries)
     end
 end
```
fhagemann, line 99: Why define a function here and call it, when you could just call the code ?

Also, is there no elegant way using `YAML.jl` to write these entries instead of using `println(io, ...)`?
Not that I have tested this, but I would expect something along the lines of


```suggestion

        # load or initialize the vector of entries
        raw     = YAML.load_file(dst, dicttype = OrderedDict)
        entries = raw isa Vector ? raw : Any[]

        ts = string(filekey.time)

        # drop any previous entry with the same timestamp
        filter!(e -> e["valid_from"] != ts, entries)

        # build a Dict containing the validity to be saved
        push!(entries, Dict{String,Any}(
            "valid_from" => ts,
            "apply" => apply,
            "category" => category,
            "mode" => mode
        ))
        
        write_from_master && @info "Write validity for $ts"

        # sort the entries by timestamp
        sort!(entries, by = e -> e["valid_from"])
        
        # create a new file with the same filename with updated entries
        YAML.write_file(dst, entries)
end
```

DaGeibl:
> @DaGeibl Thanks for the PR. Do you wanna start the review already? This sounds to me like it is still in the draft stage. Let me know as soon as it is ready to be reviewed.

We can discuss the status tomorrow in the JuLeAna call. After that a quick look to check if the logic already makes sense would be appreciated, even if there are still some bugs in the exact syntax.

src/dataprod_config.jl
DaGeibl: How important is speed here? Because currently i am only callingt` _groupings_runs` once, with your suggestion, each time `make_row` is called

theHenks:
> I fixed the failing tests. I also used ParallelProcessingTools in order to try to fix issue #120. @theHenks How can I reproduce the issue to see if i was successful with that?

You have to `read` and `write` the same `validity.yaml` file in parallel with multiple processes.

theHenks:
@DaGeibl The PR is still a draft. Is this ready for review from your side? I did not have a detailed look so far, however Felix of course reviewed a lot.

src/dataprod_config.jl
DaGeibl: Where would you like to put such a list? In the metadata?

DaGeibl:
Yeah, I would be ready for a first review. I still want to check the validity for the new datagroupings by running tests with the dataflow, so you during the review you could skip that part for now.

src/dataprod_config.jl
fhagemann: I mean, there is already runlists in the metadata, so you could just use the keys of runlists.valid for example:
https://github.com/legend-exp/legend-datasets/blob/main/runlists.yaml
Then, there would be no need of defining another list in the metadata.

Alternatively, you could define a list of Smybols just in this package. But using the info in runlists.yaml in probably better and more flexible, in case there might be more categories added in the future.

src/dataprod_config.jl
```diff
@@ -279,83 +292,127 @@ function analysis_runs end
...
 
-Get all default calibration groupings ("partitions").
-"""
-function cal_groupings_default(data::LegendData)::PropDict
-    cal_analysis_runs = PropDict()
-    for (part, runs) in data.metadata.datasets.cal_groupings.default
-        for (p, rs) in runs
-            cal_analysis_run_p = get!(cal_analysis_runs, p, DataRun[])
-            append!(cal_analysis_run_p, parse_runs(rs))
+const _cached_part_groupings = LRU{Tuple{UInt, DataCategoryLike}, StructVector{@NamedTuple{period::DataPeriod, run::DataRun}}}(maxsize = 10)
```
fhagemann, line 296: Just to be sure: `maxsize = 10` should always be enough for all cases that we might anticipate?

src/dataprod_config.jl
fhagemann: I feel like still has to be decided, but it's not me who will do so ;)

src/dataprod_config.jl
```diff
@@ -279,83 +292,127 @@ function analysis_runs end
...
+function runinfo(data::LegendData, runsel::RunCategorySelLike; runlist::Symbol = :valid)
     period, run, category = runsel
     period, run, category = DataPeriod(period), DataRun(run), DataCategory(category)
-    getproperty(runinfo(data, (period, run)), Symbol(category))
+    getproperty(runinfo(data, (period, run); runlist), Symbol(category))
 end
-runinfo(data, fk::FileKey) = runinfo(data, (fk.period, fk.run, fk.category))
-runinfo(data, selectors...) = runinfo(data, selectors)
+runinfo(data, fk::FileKey; runlist::Symbol = :valid) = runinfo(data, (fk.period, fk.run, fk.category); runlist)
+runinfo(data, selectors...; runlist::Symbol = :valid) = runinfo(data, selectors; runlist)
```
fhagemann, line 415: Instead of defining the same keyword argument default, you could just pass the keywords arguments through. 
Then, the base `runinfo` will either set the default or take whatever it was given.
This should then be explained in the `runinfo` docstring.

```suggestion
runinfo(data, fk::FileKey; kwargs...) = runinfo(data, (fk.period, fk.run, fk.category); kwargs...)
runinfo(data, selectors...; kwargs...) = runinfo(data, selectors; kwargs...)
```

src/props_db.jl
```diff
@@ -272,7 +272,11 @@ function _load_validity(validity_path::AbstractString; mode_default::AbstractStr
...
         # TODO: Cannot use readlprops here because this can be a Vector of PropDicts and not a PropDict
         #       Can readlprops be updated so that we can drop the explicit YAML dependency here?
-        entries = PropDict.(YAML.load_file(validity_path))
+        raw = ParallelProcessingTools.read_files(validity_path) do io
+            YAML.load_file(io)
+        end
+        entries = PropDict.(raw === nothing ? [] : raw)
+
         new_validity = _ValidityDict()
         for props in entries
```
fhagemann, line 281: Maybe instead of allocating an empty array, basically you just want to avoid the `for`-loop:

```suggestion
        raw = ParallelProcessingTools.read_files(validity_path) do io
            YAML.load_file(io)
        end

        new_validity = _ValidityDict()
        if !isnothing(raw)
        for props in PropDict(raw)
```
```julia
end
```

src/utils/pars_utils.jl
```diff
@@ -37,49 +37,112 @@ Write validity for a given filekey.
...
+            end
         end
-        # write validity
-        open(validity_filename, "w") do io
-            for line in sort(validity_lines)
-                println(io, line)
+
+        # load or initialize the vector of entries
+        raw     = YAML.load_file(dst, dicttype = OrderedDict)
+        entries = raw isa Vector ? raw : Any[]
```
fhagemann, line 59: can `raw` be anything else than `nothing`?
And I don't like `Any[]` at all... Looks like `entries` is a `OrderedDict{String,Any}[]`.

Does this work?
```suggestion
        raw     = YAML.load_file(dst, dicttype = OrderedDict{String, Any})
        entries::Vector{OrderedDict{String,Any}} = raw isa Vector ? raw : OrderedDict{String,Any}[]
```

fhagemann (approved):
I think this is good to be merged soon.
Some final comments from my side

src/dataprod_config.jl
fhagemann: This is still unresolved

fhagemann: In my version, runlist is still listed as argument..

src/dataprod_config.jl
fhagemann: Alright. But hardcoding stuff like this (instead of dynamically reading it from some file) will always mean that in case some new categories get added, they will not be compatible with this current version of LegendDataManagement then.

src/dataprod_config.jl
DaGeibl: In version i have this:
```


"""
    runinfo(data::LegendData)::Table
    runinfo(data::LegendData, runsel::RunSelLike)::NamedTuple
    runinfo(data::LegendData, filekey::FileKey)::NamedTuple

Get the run information for `data` based on various selection criteria.

# Arguments
- `data::LegendData`: The dataset to query run information from.

# Keyword Arguments
- `runlist::Symbol = :valid`: Specifies which runlist from `data.metadata.datasets.runlists` to use for filtering. Defaults to `:valid`.

# Returns
A table of run information with one named tuple per category (e.g. `:cal`, `:phy`), each containing `startkey`, `livetime`, and `is_analysis_run`

# Example
runinfo(data)                                   # full table of valid runs
runinfo(data, :p03)                             # all runs in period p03
runinfo(data, (:p03, :r005))                    # single-row Table for that run
runinfo(data, (:p03, :r005, :phy))              # only the :phy entry
runinfo(data, fk::FileKey)                      # same as above via FileKey
runinfo(data; runlist = :custom_list)           # use a custom runlist
"""
```

Maybe because I did not use commit suggestion but update in a bigger commit by me?

src/filekey.jl
```diff
@@ -396,6 +329,73 @@ Represents a LEGEND run selection for a specific `category`.
...
+    DataPartition(no::Int, set::Symbol = :a, cat::DataCategory = DataCategory(:cal)) = new(no, set, cat)
+end
+export DataPartition
+
+@inline DataPartition(partition::DataPartition) = partition
+Base.:(==)(a::DataPartition, b::DataPartition) = a.no == b.no && a.set == b.set && a.cat == b.cat
+Base.isless(a::DataPartition, b::DataPartition) = isless(a.no, b.no) && isless(a.set, b.set)
+
+# ToDo: Improve implementation
+Base.print(io::IO, partition::DataPartition) = print(io, "$(partition.cat.label)group$(lpad(string(partition.no), 3, string(0)))$(partition.set)")
```
theHenks, line 362: I am a little confused by the naming. Atm, we call everything a `DataPartition`, but in the metadata it is named `group`. That seems to be a little confusing for users. 
I propose to either:
- Rename `DataPartition` to `DataGroup` so that everything which was called partition is now called group
- Keep the partition name, but then also display and name everyhting something like `calpart001a` or something similar so that we just read the `groupings` files and then rename everything to partition

The naming is temporary and I am open for suggestions.

src/utils/utils.jl
```diff
@@ -10,12 +10,14 @@ using PropertyDicts
 using StructArrays
 using Unitful
 using Format
-using JSON
+using YAML
```
theHenks, line 13: Do we have to fully drop the `JSON` dependency? I thought you can still read e.g. `json` detector metadata etc after this is merged?

src/utils/management_utils.jl
```diff
@@ -48,11 +48,11 @@ get_mltrainfilename(data::LegendData, filekey::FileKey) = get_mltrainfilename(da
 
 
 
-function get_partition_channelinfo(data::LegendData, chinfo::Table, period::DataPeriodLike; unfold_partitions::Bool=false)
+function get_partition_channelinfo(data::LegendData, chinfo::Table, period::DataPeriodLike, category::DataCategoryLike; unfold_partitions::Bool=false)
     # get partition information for given period
     period = DataPeriod(period)
     # get partition information for given period and channels
-    parts = partitioninfo.(data, chinfo.channel, period)
+    parts = partitioninfo.(Ref(data), chinfo.detector, category, Ref(period))
```
theHenks, line 55: Why is the `Ref` needed here? I thought there is the broadcast specialization for `partitioninfo` and you don't need that? Or is this broken?

src/props_db.jl
theHenks, line 1: Maybe we could attach an example here to the PR for future reference. But I think the tests are already doing that.

theHenks (changes requested):
Thanks a lot @DaGeibl . I added my comments. I think I only have to general topics: 
- I would like to think about our naming of the `DataPartition` and how to lay this out
- Maybe we should introduce a new struct for the `runlists`?

Project.toml
```diff
@@ -59,6 +61,8 @@ MIMEs = "0.1, 1"
 Markdown = "<0.0.1, 1"
 Measurements = "2.2.1"
 OhMyThreads = "0.5, 0.6, 0.7, 0.8"
+OrderedCollections = "1.1"
```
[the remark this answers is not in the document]
fhagemann: I remember from SSD that `OrderedDict` did not work properly prior to `1.1`
https://github.com/JuliaCollections/OrderedCollections.jl/releases/tag/v1.1.0

`1.1`  I think added all functionality needed for `OrderedDict` to run smoothly.

src/dataprod_config.jl
```diff
@@ -253,7 +253,13 @@ end
...
-function _analysis_runs(data::LegendData, cat::DataCategoryLike)
+
+"""
+    analysis_runs(data::LegendData)
+
+Return cross-period analysis runs.
+"""
+function analysis_runs(data::LegendData, cat::DataCategoryLike)
     Table(sort(get!(_cached_analysis_runs, (objectid(data), cat)) do
         aruns::PropDict = get(data.metadata.datasets.runlists.valid, Symbol(cat), PropDict())
```
fhagemann, line 265: How would you get a list of analysis runs for a given runlist (e.g. `prl25`) then?
Any reason you are restricting the code to use `runlists.valid` here?

src/dataprod_config.jl
```diff
@@ -207,7 +207,8 @@ export get_partition_combined_periods
 
 const _cached_combined_partitions = LRU{Tuple{UInt, Symbol, Vector{Symbol}}, Vector{DataPeriod}}(maxsize = 300)
 function _get_partition_combined_periods(data::LegendData, period::DataPeriodLike, dets::Vector{<:DetectorIdLike})
```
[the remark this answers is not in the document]
fhagemann: So, ideally you would specify the types of the variables (which might not work because the variables have the same names as the input variables)
```julia
period::Symbol, dets::Vector{Symbol} = Symbol(DataPeriod(period)), Symbol.(DetectorId.(dets))
```

fhagemann:
What's the status on this?

src/props_db.jl
DaGeibl: Yes, I would also but that into the big PR, or i a new PR

src/filekey.jl
DaGeibl: I changed the Struct as discussed to output "partition" instead of group. But the Struct can read "group", "partition", or "part"

DaGeibl (commented):
I commented on the open conversations. They should be either fixed with the latest commits, or will be tackled in the big main PR,
but this PR should be good to be merged.

DaGeibl:
> What's the status on this?

I was a bit on vacation :smile:

theHenks (approved):
🤟🏻🤟🏻

PR #133, opened 2025-09-18
add dataset to LegendData and SetupConfig struct

DaGeibl:
I added the discussed functionality to add a dataset to the legend-data-config.

SetupConfig has the field `dataset` that is read from the config file. If no dataset is specified, it defaults to `:valid` as dataset. 
The config dataset can be overwritten during the construction of the LegenData struct by:
`LegendData(:setup ; dataset =  :new_dataset)`

@oschulz is this implementation fine like this?

With this added I changed the `analysis_runs` function to use the dataset specified in LegenData for the used runlist to check for analysis runs.

In the actual runinfo i did not change anything, as I understood the runinfo should always print all general available runs.
Just the is_ana_run is now dependent on the dataset.

I added a test, to check if the overwriting works properly. They are commented out for now, as I need the [PR](https://github.com/legend-exp/legend-testdata/pull/40) to be merged first

### DIFF
--- a/src/data_config.jl
+++ b/src/data_config.jl
@@ -23,6 +23,7 @@ data_path(setup, "tier", "raw", "cal", "p02", "r006", "l200-p02-r006-cal-2022122
 """
 struct SetupConfig
     paths::Vector{Pair{Vector{String}, String}}
+    dataset::String
 end
 export SetupConfig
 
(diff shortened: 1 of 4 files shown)


oschulz (changes requested):

oschulz:
Thanks @DaGeibl .

PR #141, opened 2025-12-03
Support detector name and/or channel ID based tier and file name access

jonasschlegel:
This PR enables tier file and dataset access using detector names (e.g., `V00050A`) in addition to channel IDs (e.g., `ch1084000`). This is required for the Juleana dataflow where tier outputs (`jldsp`, `jlpeaks`, `jlhit`, etc.) are stored with detector names as keys and filenames.

## Changes

**`src/legend_data.jl`**
- Add `_getindex_impl` dispatch for `Union{DetectorId, Symbol}` to build tier file paths using detector names directly
- This enables `l200.tier[:jlpeaks, filekey, DetectorId("V00050A")]` to return a path with detector name instead of channel ID

**`ext/LegendDataManagementLegendHDF5IOExt.jl`**
- Add `_resolve_channel_detector` helper function that resolves any `ChannelOrDetectorIdLike` input to a `(channel_id, detector_name)` pair via `channelinfo`
- Extend `_lh5_data_open` with fallback logic that tries multiple filename candidates in order:
  1. Channel-ID based filename (e.g., `*-ch1084000-*.lh5`)
  2. Detector-name based filename (e.g., `*-V00050A-*.lh5`)
  3. Run-level file without channel/detector in name
- Returns data from the first existing file found

## Motivation

Juleana processors follow this convention:
- **Reading**: Accept both detector names and channel IDs as input, preferring detector names
- **Writing**: Always use detector names for output filenames and HDF5 group keys

Previously, `read_ldata` and tier path lookups only supported channel-based access. With this PR, data can be accessed regardless of whether files are named by detector or channel - the reader automatically tries both naming conventions.

## Backward Compatibility

- Existing code using channel IDs continues to work unchanged
- No changes to the public API signatures
- Only adds new dispatch methods and internal fallback logic

Previously, read_ldata only supported channel-based access, requiring workarounds in processors. With this PR, data can be read regardless of whether files/datasets are named by detector or channel - the reader automatically tries both.

### DIFF
--- a/src/filekey.jl
+++ b/src/filekey.jl
@@ -731,7 +731,7 @@ _can_convert_to(::Type{DetectorId}, s) = false
 function DetectorId(s::AbstractString)
     _can_convert_to(DetectorId, s) || throw(ArgumentError("String \"$s\" does not look like a valid file LEGEND detector id"))
     length(s) < 4 && throw(ArgumentError("String \"$s\" is too short to be a valid LEGEND detector id"))
-    length(s) > 7 && throw(ArgumentError("String \"$s\" is too long to be a valid LEGEND detector id"))
+    length(s) > 9 && throw(ArgumentError("String \"$s\" is too long to be a valid LEGEND detector id"))
     DetectorId(Symbol(s))
 end
 


ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -23,11 +22,15 @@ function _is_valid_channel_or_tier(data::LegendData, rsel::Union{AnyValiditySele
...
+function _channelinfo_lookup(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, selector)
+    channelinfo(data, (rsel, selector); only_processable=false, only_usability=:all, system=:all)
+end
+
 function _get_channelid(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, det::ChannelOrDetectorIdLike)
     if LegendDataManagement._can_convert_to(ChannelId, det)
         ChannelId(det)
     elseif LegendDataManagement._can_convert_to(DetectorId, det)
-        detector2channel(data, rsel, det)
+        ChannelId(_channelinfo_lookup(data, rsel, det).channel)
```
DaGeibl: Why do you need this function you defined? why can't you just use detector2channel? To me it looks like it does the exact same. 
If you need this function it think it should be moved to the LDMUtils

DaGeibl (commented):
I only leave this two comments for now as the PR was closed

ext/LegendDataManagementLegendHDF5IOExt.jl
jonasschlegel: > Why do you need this function you defined? why can't you just use detector2channel? To me it looks like it does the exact same. If you need this function it think it should be moved to the LDMUtils

check the latest changes, sorry there was an older version I pushed.

jonasschlegel:
We will open another PR

PR #143, opened 2026-01-07
Expand LegendDataManagement documentation

DaGeibl:
Hi,
I started expanding the documentation. This process is **not** finished and this PR is currently a very early draft.

I created the pull request and the docs_update branch so that the work on the docs becomes and visible and other people could also contribute. 
Any additions, corrections, or suggestions are very welcome.


DaGeibl:
Hi, quite a while ago I started this PR and expanded a bit the docs. I only started this a draft in the hoped to find some to extend this. Sadly this did not happen yet.
So i thought let's just merge this, as it is better than nothing. Otherwise this would remain a draft PR for quite much longer.

PR #157, opened 2026-03-10
Support extras (crack, topgroove, bottom_cylinder, top_cylinder) in SSD config file creation

giriPHM:
… Refactored and updated from @apearsonn's work in PR #81. Integrated logic into the modern extension structure and verified compatibility with the current metadata pipeline. I did run the tests as well. All the tests passed. There was only one warning that the test seemed to generate. I have attached the screenshot for the output from the test below.
Screenshot 2026-03-09 at 1 22 39 PM

### DIFF
--- a/ext/LegendDataManagementSolidStateDetectorsExt.jl
+++ b/ext/LegendDataManagementSolidStateDetectorsExt.jl
@@ -273,12 +273,14 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
     has_top_groove = has_extras && haskey(geo.extra, :topgroove)
     has_top_cylinder = has_extras && haskey(geo.extra, :top_cylinder) && geo.extra.top_cylinder.radius_in_mm != crystal_radius
     has_crack = has_extras && haskey(geo.extra, :crack)
-    @assert !(has_top_cylinder && has_bottom_cylinder) "Having both a top and bottom cylinder is not supported."
+    if has_top_cylinder && has_bottom_cylinder
+        throw(ArgumentError("Having both a top and bottom cylinder is not supported."))
+    end
 
     bottom_cyl_radius = has_bottom_cylinder ? geo.extra.bottom_cylinder.radius_in_mm : crystal_radius 
     top_cyl_radius =  has_top_cylinder ? geo.extra.top_cylinder.radius_in_mm : crystal_radius 
 
-    world_radius = 1.2 * maximum([crystal_radius, top_cyl_radius, bottom_cyl_radius])
+    world_radius = 1.2 * max(crystal_radius, top_cyl_radius, bottom_cyl_radius)
 
     config_dict = dicttype(
         "name" => diode_meta.name,
@@ -368,7 +370,7 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
         has_borehole = hasproperty(geo, :borehole)
         borehole_radius = 0
         if is_coax && !has_borehole
-            error("Coax detectors should have boreholes")
+            throw(ArgumentError("Coax detectors should have boreholes"))
         end
         if has_borehole
             borehole_depth = geo.borehole.depth_in_mm
@@ -402,15 +404,17 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
                 borehole_taper_angle = geo.taper.borehole.angle_in_deg
                 borehole_taper_radius = borehole_taper_height * tand(borehole_taper_angle)
             else
-                error("The borehole taper needs either radius_in_mm or angle_in_deg")
+                throw(ArgumentError("The borehole taper needs either radius_in_mm or angle_in_deg"))
             end
             has_borehole_taper = borehole_taper_height > 0 && borehole_taper_angle > 0
-            @assert !(has_borehole_taper && has_top_groove) "Having both a borehole taper and a top groove is not supported."
+            if has_borehole_taper && has_top_groove
+                throw(ArgumentError("Having both a borehole taper and a top groove is not supported."))
+            end
             if has_borehole_taper && !has_borehole 
-                error("A detector without a borehole cannot have a borehole taper.")
+                throw(ArgumentError("A detector without a borehole cannot have a borehole taper."))
             end
             if has_borehole_taper && is_coax
-                error("Coax detectors should not have borehole tapers")
+                throw(ArgumentError("Coax detectors should not have borehole tapers"))
             end
             if has_borehole_taper
                 hZ = borehole_taper_height + 2*gap
@@ -441,11 +445,11 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
                 top_taper_angle = geo.taper.top.angle_in_deg
                 top_taper_radius = top_taper_height * tand(top_taper_angle)
             else
-                error("The top taper needs either radius_in_mm or angle_in_deg")
+                throw(ArgumentError("The top taper needs either radius_in_mm or angle_in_deg"))
             end
             has_top_taper = top_taper_height > 0 && top_taper_angle > 0
             if has_top_taper
-                r_center = r_center = (has_top_cylinder ? top_cyl_radius : crystal_radius) - top_taper_radius / 2
+                r_center = (has_top_cylinder ? top_cyl_radius : crystal_radius) - top_taper_radius / 2
                 hZ = top_taper_height/2 + 1gap
                 h = 2 * hZ
                 Δr = hZ * tand(top_taper_angle)         
@@ -482,7 +486,7 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
                 bot_taper_angle = geo.taper.bottom.angle_in_deg
                 bot_taper_radius = bot_taper_height * tand(bot_taper_angle)
             else
-                error("The bottom outer tape needs either radius_in_mm or angle_in_deg")
+                throw(ArgumentError("The bottom outer taper needs either `radius_in_mm` or `angle_in_deg`"))
             end
             has_bot_taper = bot_taper_height > 0 && bot_taper_angle > 0
             if has_bot_taper
@@ -642,7 +646,7 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
             top_borehole_radius = is_coax ? 0 : borehole_radius
             r_in = has_top_groove ? top_groove_radius : top_borehole_radius
             r_out = top_cyl_radius
-            if has_borehole_taper r_in += borehole_taper_radius end # but no top groove, see assert above
+            if has_borehole_taper r_in += borehole_taper_radius end # but no top groove, see throw(ArgumentError(...)) above
             if has_top_taper r_out -= top_taper_radius end
             push!(mantle_contact_parts, dicttype(
                     "tube" => dicttype(
@@ -673,7 +677,7 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
                             "from" => borehole_radius,
                             "to" => top_groove_radius + li_thickness 
                         ),
-                        "h" =>li_thickness,
+                        "h" => li_thickness,
                         "origin" => [0, 0, crystal_height - top_groove_height - li_thickness / 2]
                     )
                 )
--- a/ext/LegendDataManagementSolidStateDetectorsExt.jl
+++ b/ext/LegendDataManagementSolidStateDetectorsExt.jl
@@ -771,8 +771,72 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
                 )
             )
         end
-        
         mantle_contact_parts
+    end
+
+    if haskey(diode_meta.geometry, :extra) 
+        if haskey(diode_meta.geometry.extra, :crack)
+            
+            # cut the crack volume from the semiconductor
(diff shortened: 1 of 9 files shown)


hervasa2 (changes requested):
Thanks for the initiative! There are many changes that undo recent updates which have nothing to do with the additional geometry support. Please undo these and address the ignoring the Li thickness comment. My recommendation is use something with exaggerated `n_thickness` and plot to check like so:
```julia
sim = Simulation{T}(LegendData, meta, xtal_meta, verbose = true, HPGeEnvironment("LAr", 87u"K"), n_thickness = 5u"mm");
plot(sim.detector, st = :slice)
```
Could you also please include a test of each new feature, showing a snippet of the metadata and the resulting plot (using exaggerated `n_thickness`). Also test one detector with no "extras" to see that the old functionality still works.

hervasa2 (changes requested):

fhagemann:
This should also be extended to include support for `top_cylinder` for `V06649M` (see #159 and https://github.com/legend-exp/legend-detectors/pull/74)

giriPHM:
I have used 
```
sim_2 = Simulation{T}(LegendData, meta, xtal_meta, verbose = true, HPGeEnvironment("LAr", 87u"K"), n_thickness = 2u"mm");
plot(sim_2.detector, st = :slice, legend=:topright, title="$det")
```
The top taper for V02160A and V02162B might look a bit weird, but that is just because the angle is small and the height of the taper is scaled by tangent of the angle.

giriPHM:
Since the n+ contact geometry of V02162B (one with the top groove) looked a bit weird, I went back and made some fixes to the geometry. Now, I think it looks good. Since the top groove height is just 2.7 mm, I have used n_thickness of 1.5 mm to see things better.

theHenks:
@giriPHM Thank you for your work here. Are you certain that all this metadata is public? Since we are sharing this in a public repository, I recommend double-checking what can be shown and what cannot to ensure we do not release any confidential detector data.

giriPHM:
> @giriPHM Thank you for your work here. Are you certain that all this metadata is public? Since we are sharing this in a public repository, I recommend double-checking what can be shown and what cannot to ensure we do not release any confidential detector data.

That is a good point, Florian. I was posting those snippets in response to David's comment regarding the test of the features. But I will remove that information till we figure this issue out!

giriPHM:
@Copilot. Can you suggest some tests for the added features to address the patch coverage?

giriPHM:
Add top cylinder feature with tests; fix borehole depth parameter in active_volume.jl

- Implemented top cylinder as a new geometry feature
- Added corresponding tests for extra features functionality
- Fixed bug in active_volume.jl where borehole depth was incorrectly using radius parameter

hervasa2:
- Refactored `mantle_contact` implementation to include extras and remove redundancies. The result is much more compact and maintainable.
- Taper mantles are now handled correctly with a dedicated function.
- Added tests for each `extra`. This involves creating one base detector and removing the different extras from the base volume. A point from the removed volume is tested to be in the original detector but outside of the modified volume. An additional test is added to check that the active volume of the modified volume is smaller than the base detector active volume.

- [x] **Base:**

Screenshot 2026-04-29 at 16 02 43

- [x] **Top cylinder:**
Screenshot 2026-04-29 at 14 43 03

- [x] **Bottom cylinder:**
Screenshot 2026-04-29 at 14 43 44

- [x] **Top groove:**
Screenshot 2026-04-29 at 14 47 03

- [x] **Crack:**
Screenshot 2026-04-29 at 14 47 37

fhagemann:
In order to review this conceptually (without looking at every line of code): can you also attach picture for the „normal“ mantle taper case. Just to make sure that the refactoring did not break anything? :)

hervasa2:
See exaggerated "normal" mantle taper cases without extras:

- [x] IC
image

- [x] COAX
image

- [x] PPC
image

- [x] BEGe
image

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -180,13 +260,25 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
...
+    has_bottom_cylinder = has_extras && haskey(geo.extra, :bottom_cylinder) && geo.extra.bottom_cylinder.radius_in_mm != crystal_radius
+    has_top_groove = has_extras && haskey(geo.extra, :topgroove)
+    has_top_cylinder = has_extras && haskey(geo.extra, :top_cylinder) && geo.extra.top_cylinder.radius_in_mm != crystal_radius
+    has_crack = has_extras && haskey(geo.extra, :crack)
+    @assert !(has_top_cylinder && has_bottom_cylinder) "Having both a top and bottom cylinder is not supported."
+
+    bottom_cyl_radius = has_bottom_cylinder ? geo.extra.bottom_cylinder.radius_in_mm : crystal_radius 
+    top_cyl_radius =  has_top_cylinder ? geo.extra.top_cylinder.radius_in_mm : crystal_radius 
+
+    world_radius = 1.2 * maximum([crystal_radius, top_cyl_radius, bottom_cyl_radius])
```
fhagemann: This will avoid creating/allocating a vector:
```suggestion
    world_radius = 1.2 * max(crystal_radius, top_cyl_radius, bottom_cyl_radius)
```

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -232,21 +324,55 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
         ))
     
     # main crystal
+    bottom_cyl_height = if has_bottom_cylinder
```
fhagemann, line 329: Is this intended to call this `bottom_cyl_height`?

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -232,21 +324,55 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
...
+             crystal_height
+        end
+    end
     semiconductor_geometry_basis = dicttype("cone" => dicttype(
-        "r" => crystal_radius,
-        "h" => crystal_height,
-        "origin" => [0, 0, crystal_height / 2]
+        "r" => bottom_cyl_radius,
+        "h" => bottom_cyl_height,
+        "origin" => [0, 0, bottom_cyl_height / 2]
```
fhagemann, line 341: I'm a bit confused why you would call this `bottom_XXX` while most of the time it will just be the actual `radius` and `height` (?)

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -300,20 +429,23 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
...
+            elseif hasproperty(geo.taper.top, :angle_in_deg)
+                top_taper_angle = geo.taper.top.angle_in_deg
                 top_taper_radius = top_taper_height * tand(top_taper_angle)
             else
                 error("The top taper needs either radius_in_mm or angle_in_deg")
             end
             has_top_taper = top_taper_height > 0 && top_taper_angle > 0
             if has_top_taper
-                r_center = crystal_radius - top_taper_radius / 2
+                r_center = r_center = (has_top_cylinder ? top_cyl_radius : crystal_radius) - top_taper_radius / 2
```
fhagemann: `r_center` is duplicated here:

```suggestion
                r_center = (has_top_cylinder ? top_cyl_radius : crystal_radius) - top_taper_radius / 2
```

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -264,19 +390,22 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
...
+                borehole_taper_angle = geo.taper.borehole.angle_in_deg
                 borehole_taper_radius = borehole_taper_height * tand(borehole_taper_angle)
             else
                 error("The borehole taper needs either radius_in_mm or angle_in_deg")
             end
             has_borehole_taper = borehole_taper_height > 0 && borehole_taper_angle > 0
+            @assert !(has_borehole_taper && has_top_groove) "Having both a borehole taper and a top groove is not supported."
             if has_borehole_taper && !has_borehole 
                 error("A detector without a borehole cannot have a borehole taper.")
             end
```
fhagemann: I think this is more of a code-style question, but can we decide for either `@assert` or `error` syntax for this whole file. This would make it easier to search for all errors by Ctrl+F on either `@assert` or `error`.
Maybe even make this more explicit an `ArgumentError`: `throw(ArgumentError(...))`?

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -338,46 +470,50 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
...
-                            "to" => r_out
+            bot_taper_height = geo.taper.bottom.height_in_mm
+            if hasproperty(geo.taper.bottom, :radius_in_mm)
+                bot_taper_radius = geo.taper.bottom.radius_in_mm
+                bot_taper_angle = atand(bot_taper_radius, bot_taper_height)
+            elseif hasproperty(geo.taper.bottom, :angle_in_deg)
+                bot_taper_angle = geo.taper.bottom.angle_in_deg
+                bot_taper_radius = bot_taper_height * tand(bot_taper_angle)
+            else
+                error("The bottom outer tape needs either radius_in_mm or angle_in_deg")
```
fhagemann:
```suggestion
                error("The bottom outer taper needs either `radius_in_mm` or `angle_in_deg`")
```

ext/LegendDataManagementSolidStateDetectorsExt.jl
```diff
@@ -491,180 +636,211 @@ function create_SSD_config_dict_from_LEGEND_metadata(diode_meta::PropDict, xtal_
...
-                        "to" => r_bot
+                )
+            )
+            push!(mantle_contact_parts, dicttype(
+                    "tube" => dicttype(
+                        "r" => dicttype(
+                            "from" => borehole_radius,
+                            "to" => top_groove_radius + li_thickness 
+                        ),
+                        "h" =>li_thickness,
```
fhagemann:
```suggestion
                        "h" => li_thickness,
```

test/test_config_files/V00000A.yaml
```diff
@@ -0,0 +1,35 @@
...
+  taper:
+    top:
+      angle_in_deg: 0
+      height_in_mm: 0
+    bottom:
+      angle_in_deg: 0
+      height_in_mm: 0
+    borehole:
+      angle_in_deg: 0
+      height_in_mm: 0
```
fhagemann: Just for the sake of testing: do we need all of these `tapers`, especially if all the values are zero? If not, I would remove them (or just keep them for one detector maybe, to check that the tapers are ignored) to improve readability..

test/test_ext_ssd.jl
fhagemann, line 1: Optional: do we want to test that all the errors are thrown (existence of both top and bottom cylinder for example)?

ext/LegendDataManagementSolidStateDetectorsExt.jl
hervasa2: To include the extras without many redundancies the logic here is that all detectors have a bottom crystal. This way you move the if statement to the radius and height definition rather than defining a whole new geometry dict.

ext/LegendDataManagementSolidStateDetectorsExt.jl
hervasa2: Everything has been made consistent with `throw(ArgumentError(...))`

test/test_config_files/V00000A.yaml
hervasa2: All zero-valued tapers have been removed from tests. This caused `get_active_volume` to fail -> patched with `get`

test/test_ext_ssd.jl
hervasa2: I think this is a bit of overkill

src/active_volume.jl
```diff
@@ -126,20 +126,20 @@ function get_active_volume(pd::PropDict, ::Val{:icpc}, fccd::T = .0) where {T <:
...
+    h = get(g.taper.bottom, :height_in_mm, 0) + fccd * (1 - sin(α) - cos(α))/sin(α)
     x = h * tan(α)
     taper_bottom_volume = iszero(x) || !isfinite(x) ? zero(T) : get_outer_taper_volume(R, R - x, h)
     
     # Borehole taper
-    α = g.taper.bottom.angle_in_deg / 360 * 2π
-    h = g.taper.bottom.height_in_mm + fccd * (1 - sin(α) - cos(α))/sin(α)
+    α = get(g.taper.borehole, :angle_in_deg, 0)/ 360 * 2π
+    h = get(g.taper.borehole, :height_in_mm, 0) + fccd * (1 - sin(α) - cos(α))/sin(α)
     x = h * tan(α)
```
fhagemann, line 143: If  `α` is 0, then `sin(α)` is also 0, and h will be `Inf` or `NaN` (depending on `fccd`). Is this problematic is any way?

fhagemann (commented):
Looks good! One final comment

test/test_ext_ssd.jl
fhagemann: Alright :)

src/active_volume.jl
hervasa2: This is caught by `!isfinite(x)` further down which is true for any div by zero.

fhagemann (approved):

fhagemann:
I guess someone (@theHenks / @oschulz ) can merge this now :)

PR #163, opened 2026-03-25
Remove indent argument from writelprops

jonasschlegel:
Drop the indent keyword argument from writelprops. This is not compatible with YAML (only indent::Int=4), but you can completely leave it out as @oschulz suggested.


theHenks (changes requested):

DaGeibl (approved):
I reworked the indent logic. It should be fine to be merged now.

theHenks (changes requested):

theHenks (approved):

PR #166, opened 2026-04-10
Change to detector ID based jlevt reading in read_ldata

jonasschlegel:
In [calibrate_all l48](https://github.com/legend-exp/LegendEventAnalysis.jl/blob/main/src/calibrate_all.jl)  we defined `trig_e_det` instead of the old `trig_e_ch` version, which was not yet fixed in the `read_ldata` access via `read_ldata(l200, :jlevt, fk_evt, det))`. I think we then also have to reprocess the PRL data set at some point. I dont want to allow both options for `trig_e_det` and `trig_e_ch`. But read_ldata works for the latest version of `LegendEventAnalysis` with `det` and `ch` argument.


jonasschlegel:
reopen in another PR

PR #170, opened 2026-04-20
change key filter of raw files to only allow DetectorID's

### LINKED ISSUE #135: Bug Fix `read_ldata`
After the successful merge of the `YAML` update, the `read_ldata` function seems to be broken, given the current new `LH5` data layout. In particular, there is an inconsistency between the access of individual detectors and the mapping within the `lh5` files.

DaGeibl:
Hi this PR fixes https://github.com/legend-exp/LegendDataManagement.jl/issues/135. 
I described the Problem in greater detail in that issue. 

This fix changes the filtering of keys for the raw tier to only allow DetectorID's.

I would also change the warning line, as it currently is very noisy when opening raw files. Should i instead replace it with an `@debug` or remove the entire line?

### DIFF
--- a/ext/LegendDataManagementLegendHDF5IOExt.jl
+++ b/ext/LegendDataManagementLegendHDF5IOExt.jl
@@ -15,7 +15,7 @@ const AbstractDataSelectorLike = Union{AbstractString, Symbol, DataTierLike, Dat
 const PossibleDataSelectors = [DataTier, DataCategory, DataPeriod, DataRun, DataPartition, ChannelId, DetectorId]
 
 function _is_valid_id_or_tier(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, id::ChannelOrDetectorIdLike)
-    if LegendDataManagement._can_convert_to(ChannelId, id) ||  LegendDataManagement._can_convert_to(DetectorId, id) ||  LegendDataManagement._can_convert_to(DataTier, id)
+    if LegendDataManagement._can_convert_to(DetectorId, id)
         true  
     else  
         @warn "Skipped $id since it is neither a valid `ChannelId`, `DetectorId` nor a `DataTier`"  


oschulz:
Is kind of a workaround for the dummy-channel problem? If so, we expect that to go away in future data productions (maybe)? Can we just handle this dummy-channel explicitly in that conditional statement?

theHenks (changes requested):
In general, this is good. Some feedback:
- There is some more legacy functionality in there which uses `ChannelId`. Let's remove everything.
- Change warning to `@debug` as you proposed
- More tests. I would prefer tests that cover the current problem.

DaGeibl:
Hi, the reason for this PR is not only the Dummy channels, but the current key structure of raw files that look like this:

`"B00000C", ..., "B00079B", "B00079C", "BF862-01", ..., "BF862-21", "BF862-22", "BF862-23", "BF862-24", "BSLN01", "MUON01", "PMT101", "PMT102", ..., "PMT709", "PMT710", "PULS01", "PULS01ANA", "S002", ... , "S099", "S100", "V00048A", "V00048B", ...,  "V11925A", "V11947B", "V13044A", "V13046A", "V13049A", "V14618A", "V14654A", "V14673A", "ch1027200", ...,  "ch2004822", "extra", "raw"
`
So it has the `DetectorID` keys that link to data but also channelID's as well as this `raw` key, that has in the layer below again `DetectorID` keys.
with `readldata` we try to open each key after checking if it is valid:

```
function LegendDataManagement.read_ldata(f::Base.Callable, data::LegendData, rsel::Tuple{DataTierLike, FileKey}; kwargs...)
    ids = _lh5_data_open(data, rsel[1], rsel[2], "") do h
        keys(h)
    end
    ids = filter(x -> _is_valid_id_or_tier(data, rsel[2], x), ids)
    ...
```

For raw files we only want to open `DetectorID`'s and none of the other additional keys. 
What I did to fix the function for raw files, is to only allow `DetectorID` keys. But I now realized this does not work, because for the other tiers we also use `DataTier` as key. So we need to find sth else

So I would propose one of these fixes:

- write our own raw file without these keys
- have a special case for raw files that only allows DetectorID's (e.g. if DataTierLke == :raw do ...)

DaGeibl:
Hi, I will close this PR as @jonasschlegel will fix the problem properly in his [PR](https://github.com/legend-exp/LegendDataManagement.jl/pull/172)

PR #179, opened 2026-06-25
Add allow_cylindrical_asymmetry

hervasa2:
## Summary
- Add `allow_cylindrical_asymmetry` keyword. Default is set to `true`:

Whether to allow geometry features that break the cylindrical (azimuthal) symmetry of the detector, such as a `crack` defined in the metadata `geometry.extra`. When `true` (default) and such a feature is present, the affected volume is cut from the semiconductor and contacts and the simulation grid is switched to a full 3D azimuthal range (`phi` from `0` to `360`). When `false`, these features are ignored and the simulation remains 2D (azimuthally symmetric).

- Add docstring entries for the `allow_cylindrical_asymmetry` keyword to the `SolidStateDetector` and `Simulation` constructors.
- Add a `"Test allow_cylindrical_asymmetry"` testset that verifies a `crack` in the metadata switches the simulation grid to a full 3D azimuthal range (`sim.config_dict["grid"]["axes"]["phi"]["to"] == 360`) when asymmetry is allowed (default), and is ignored (`phi["to"] == 0`) when disabled.


gipert:
@theHenks could you merge this and release a new version? we need this for legend-simflow

theHenks (approved):

PR #180, opened 2026-06-26
read_ldata: use DetectorId only (drop ChannelId from the public API)

jonasschlegel:
## What didn't work before
`read_ldata` accepted **both** a `ChannelId` and a `DetectorId` for the detector
argument. The two are just different names for the same physical detector, so
every call path had to convert back and forth (`detector2channel` /
`channel2detector`) and then re-check that the two agreed — including a
`@assert … "Channel mismatch"` that could fail.

That dual API is redundant and a source of subtle bugs: the conversions and the
consistency check add complexity without giving the caller anything they can't
already express with a `DetectorId`. Everything `read_ldata` needs downstream
(`partitioninfo`, the event-tier `trig_e_det` column, …) already works with a
`DetectorId` directly.

## What works now
`read_ldata` takes a `DetectorId` everywhere; the channel↔detector helpers and the
consistency assert are removed (−45 lines net). The event-tier filter now matches
the detector directly via `trig_e_det` instead of first converting it to a channel
number.

## Stack
Foundation of the `read_ldata` refactor. Built on, in order, by: **internals
refactor** → **per-detector event-tier slicing** → **cross-tier `filtertier`**.
(Docs/tests and the thread-safety fix are independent.)

### DIFF
--- a/ext/LegendDataManagementLegendHDF5IOExt.jl
+++ b/ext/LegendDataManagementLegendHDF5IOExt.jl
@@ -2,7 +2,6 @@ module LegendDataManagementLegendHDF5IOExt
 
 using LegendDataManagement
 LegendDataManagement._lh5_ext_loaded(::Val{true}) = true
-using LegendDataManagement.LDMUtils: detector2channel, channel2detector
 using LegendDataManagement: RunCategorySelLike
 using LegendHDF5IO
 using LegendDataTypes: fast_flatten, flatten_by_key
@@ -10,38 +9,8 @@ using StructArrays
 using TypedTables, PropertyFunctions
 using Distributed, ProgressMeter
 
-const ChannelOrDetectorIdLike = Union{ChannelIdLike, DetectorIdLike}
-const AbstractDataSelectorLike = Union{AbstractString, Symbol, DataTierLike, DataCategoryLike, DataPeriodLike, DataRunLike, DataPartitionLike, ChannelOrDetectorIdLike}
-const PossibleDataSelectors = [DataTier, DataCategory, DataPeriod, DataRun, DataPartition, ChannelId, DetectorId]
-
-function _is_valid_id_or_tier(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, id::ChannelOrDetectorIdLike)
-    if LegendDataManagement._can_convert_to(ChannelId, id) ||  LegendDataManagement._can_convert_to(DetectorId, id) ||  LegendDataManagement._can_convert_to(DataTier, id)
-        true  
-    else  
-        @warn "Skipped $id since it is neither a valid `ChannelId`, `DetectorId` nor a `DataTier`"  
-        false  
-    end  
-end
-
-function _get_channelid(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, det::ChannelOrDetectorIdLike)
-    if LegendDataManagement._can_convert_to(ChannelId, det)
-        ChannelId(det)
-    elseif LegendDataManagement._can_convert_to(DetectorId, det)
-        detector2channel(data, rsel, det)
-    else
-        throw(ArgumentError("$det is neither a ChannelId nor a DetectorId"))
-    end
-end
-
-function _get_detectorid(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, det::ChannelOrDetectorIdLike)
-    if LegendDataManagement._can_convert_to(DetectorId, det)
-        DetectorId(det)
-    elseif LegendDataManagement._can_convert_to(ChannelId, det)
-        channel2detector(data, rsel, det)
-    else
-        throw(ArgumentError("$det is neither a ChannelId nor a DetectorId"))
-    end
-end
+const AbstractDataSelectorLike = Union{AbstractString, Symbol, DataTierLike, DataCategoryLike, DataPeriodLike, DataRunLike, DataPartitionLike, DetectorIdLike}
+const PossibleDataSelectors = [DataTier, DataCategory, DataPeriod, DataRun, DataPartition, DetectorId]
 
 
 const dataselector_bytypes = Dict{Type, String}()
@@ -146,7 +115,6 @@ function __init__()
     (@isdefined DataCategory) && extend_datatype_dict(DataCategory, "datacategory")
     (@isdefined Timestamp) && extend_datatype_dict(Timestamp, "timestamp")
     (@isdefined FileKey) && extend_datatype_dict(FileKey, "filekey")
-    (@isdefined ChannelId) && extend_datatype_dict(ChannelId, "channelid")
     (@isdefined DetectorId) && extend_datatype_dict(DetectorId, "detectorid")
     (@isdefined DataPartition) && extend_datatype_dict(DataPartition, "datapartition")
 end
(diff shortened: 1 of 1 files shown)


ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -2,46 +2,15 @@ module LegendDataManagementLegendHDF5IOExt
...
-
-function _get_detectorid(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, det::ChannelOrDetectorIdLike)
-    if LegendDataManagement._can_convert_to(DetectorId, det)
-        DetectorId(det)
-    elseif LegendDataManagement._can_convert_to(ChannelId, det)
-        channel2detector(data, rsel, det)
-    else
-        throw(ArgumentError("$det is neither a ChannelId nor a DetectorId"))
-    end
-end
```
DaGeibl, line 44: Dropping _get_detectorid will break [this line](https://github.com/legend-exp/LegendMakie.jl/blob/05550eb75a9bb2a9c97ba90a3385b7da8ba5e9ad/ext/LegendMakieLegendDataManagementExt.jl#L250) in LegendMakie as it relies on `_get_detectorid`. We can also drop the dual use of ChannelId an DetectorId in LegendMakie, but we need PR also in changing this there.

DaGeibl: I fixed this in https://github.com/legend-exp/LegendMakie.jl/pull/69

DaGeibl:
Hi,
I just noticed all the tests are failing. I think the best workflow would be:

1. open a PR fixing https://github.com/legend-exp/LegendDataManagement.jl/issues/190 so that we have a clean wokring state for this package
2. rebase all new features on the merged bug fix
3. review all new PR's

jonasschlegel:
Tests will fail before [https://github.com/legend-exp/LegendDataManagement.jl/pull/192](url) is merged, but ready to review @theHenks @DaGeibl from my side (first PR in the read_ldata stack)

jonasschlegel:
> Hi, I just noticed all the tests are failing. I think the best workflow would be:
> 
> 1. open a PR fixing [Error regarding Property functions #190](https://github.com/legend-exp/LegendDataManagement.jl/issues/190) so that we have a clean wokring state for this package
> 2. rebase all new features on the merged bug fix
> 3. review all new PR's

I opened a PR on that: [https://github.com/legend-exp/LegendDataManagement.jl/pull/192](url)

jonasschlegel:
test will be added in [https://github.com/legend-exp/LegendDataManagement.jl/pull/195](url) (draft)

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -237,22 +201,23 @@ function LegendDataManagement.read_ldata(f::Base.Callable, data::LegendData, rse
     ids = _lh5_data_open(data, rsel[1], rsel[2], "") do h
         keys(h)
     end
-    ids = filter(x -> _is_valid_id_or_tier(data, rsel[2], x), ids)
+    # Top-level keys are either the tier itself (event-tier files) or detector ids.
+    valid(x) = LegendDataManagement._can_convert_to(DetectorId, x) ||
+               LegendDataManagement._can_convert_to(DataTier, x)
+    ids = filter(valid, ids)
```
DaGeibl, line 206: in the `_is_valid_id_or_tier` function this replaces we threw a warning if x is neither a tier or detector. This could be helpful debugging. Do we maybe want to keep that?

DaGeibl (approved):
These changes look fine i have only a minor comment.

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -146,7 +115,6 @@ function __init__()
     (@isdefined DataCategory) && extend_datatype_dict(DataCategory, "datacategory")
     (@isdefined Timestamp) && extend_datatype_dict(Timestamp, "timestamp")
     (@isdefined FileKey) && extend_datatype_dict(FileKey, "filekey")
-    (@isdefined ChannelId) && extend_datatype_dict(ChannelId, "channelid")
```
theHenks, line 149: This remove the option to write generic `ChannelId` to disk.

theHenks (changes requested):
What is with the tests after this change?

ext/LegendDataManagementLegendHDF5IOExt.jl
jonasschlegel: Yes, but do we still need that in any case?

theHenks: yes

ext/LegendDataManagementLegendHDF5IOExt.jl
jonasschlegel: So the idea of this function is to skip keys in the format "chxxx" and I would not add a warning for that because thats what the function is written for. A valid DetId is `tier_expr = r"^([a-z]+)$"` and a valid `DataTier` is any lowercase letter. So if we keep `DetectorId` and `DataTier` format up to date that should be fine. The only warning is usefull when the data only consists of unvalid keys and an empty `NamedTuple()` is returned, so we can add something like: `isempty(ids) && throw(ArgumentError("No detector or tier group found in $(basename(data.tier[rsel[1], rsel[2]]))"))`

ext/LegendDataManagementLegendHDF5IOExt.jl
jonasschlegel: Ok do you suggest to keep it or what shall we do?

theHenks: This functionality has nothing to do with `read_ldata`. This is the actual `LegendHDF5IO` extension which defines the custom generic datatype serializations for our LEGEND structs such as `DetectorId`, `ChannelId` etc.
Removing this will prevent operations such as 
``` julia
lh5open("chids.lh5", "cw") do ds
   ds["chids"] = ChannelId.(rand(1:100, 10))
end
```

ext/LegendDataManagementLegendHDF5IOExt.jl
DaGeibl: I mean this should not only skip channelID keys, but all keys, that are neither DetectorID nor a tier. I think in the raw tier we also have that case. but i guess it can be noise if you also have all the channelID keys still present in the raw tier. so i guess we can drop the warning

jonasschlegel: I added a warning

ext/LegendDataManagementLegendHDF5IOExt.jl
jonasschlegel: done.

jonasschlegel: Tests will be added afterwards

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -2,46 +2,15 @@ module LegendDataManagementLegendHDF5IOExt
...
-function _get_detectorid(data::LegendData, rsel::Union{AnyValiditySelection, RunCategorySelLike}, det::ChannelOrDetectorIdLike)
-    if LegendDataManagement._can_convert_to(DetectorId, det)
-        DetectorId(det)
-    elseif LegendDataManagement._can_convert_to(ChannelId, det)
-        channel2detector(data, rsel, det)
-    else
-        throw(ArgumentError("$det is neither a ChannelId nor a DetectorId"))
-    end
-end
+const AbstractDataSelectorLike = Union{AbstractString, Symbol, DataTierLike, DataCategoryLike, DataPeriodLike, DataRunLike, DataPartitionLike, DetectorIdLike}
```
theHenks: Are those actually used?

```suggestion

```

ext/LegendDataManagementLegendHDF5IOExt.jl
```diff
@@ -2,46 +2,15 @@ module LegendDataManagementLegendHDF5IOExt
...
+const PossibleDataSelectors = [DataTier, DataCategory, DataPeriod, DataRun, DataPartition, DetectorId]
```
theHenks:
```suggestion
```

theHenks: Same here

PR #181, opened 2026-06-26
Refactor read_ldata internals: tier classes, path resolver, apply helper

jonasschlegel:
## What didn't work before
Inside a tier file, a detector's data can live at one of two paths: the old
`<det>/<tier>` or the new `<tier>/<det>`. `read_ldata` had this path **hardcoded**
to the old `<det>/<tier>`, so it simply couldn't find data written in the new
`<tier>/<det>` layout (`jldsp`, `jlhit`).

Two more rough edges:
- When you ask for a whole tier *without* naming a detector
  (`read_ldata(l200, :jldsp, fk)`), it only looked at the **top level** of the
  file. For a `<tier>/<det>` file that means it saw the `jldsp` folder but not the
  detectors inside it.
- The "select columns / apply a filter / return as-is" logic was copy-pasted
  inline inside the main read function.

We're moving **all** tiers to the `<tier>/<det>` layout — for consistency, and
because reading through `<tier>/<det>` (e.g. `raw/<det>`) is measurably faster than
`<det>/<tier>` (writer side: the Juleana PR that switches tier outputs to
`<tier>/<det>`). So `read_ldata` has to (a) find the data in **either** layout,
(b) list detectors that now live one level down, and (c) expose a single read path
that the upcoming event-tier code can reuse.

- A small resolver tries the paths **in order** — `<tier>/<det>` (new) → `<det>/<tier>`
  (old) → the bare `<tier>` (for single-detector files) — so both layouts just work.
- `read_ldata` decides which **file** to open from the tier type (some tiers are one
  file per detector, others one shared file), instead of guessing by probing files.
- Asking for a tier without a detector now looks **inside** the tier folder and
  returns one table per detector.
- All the read/filter logic lives in one helper now; the filter also tolerates
  `missing` (those rows are dropped). `subgroup=:dataQC` lets you reach a sub-table
  (e.g. inside `jlhit`).
- Bug guard: if a detector isn't present, `read_ldata` no longer silently returns
  **all** detectors — it respects `ignore_missing` / raises a clear error.

## Stack
On top of **PR1** (DetectorId-only API). Writer-side counterpart: the Juleana PR
that moves tier outputs to `<tier>/<det>`. Required by **PR3** (per-detector
event-tier slicing) and **PR4** (cross-tier `filtertier`).

### DIFF
--- a/ext/LegendDataManagementLegendHDF5IOExt.jl
+++ b/ext/LegendDataManagementLegendHDF5IOExt.jl
@@ -119,99 +119,107 @@ function __init__()
     (@isdefined DataPartition) && extend_datatype_dict(DataPartition, "datapartition")
 end
 
-function _lh5_data_open(f::Function, data::LegendData, tier::DataTierLike, filekey::FileKey, det::DetectorIdLike, mode::AbstractString="r")
-    det_filename = data.tier[DataTier(tier), filekey, det]
-    filename = data.tier[DataTier(tier), filekey]
-    if isfile(det_filename)
-        @debug "Read from $(basename(det_filename))"
-        LegendHDF5IO.lh5open(f, det_filename, mode)
-    elseif isfile(filename)
-        @debug "Read from $(basename(filename))"
-        LegendHDF5IO.lh5open(f, filename, mode)
+const _evt_tiers = DataTier.([:jlevt, :jlskm])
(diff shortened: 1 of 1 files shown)


jonasschlegel:
@theHenks @DaGeibl ready to review, but we need to merge [https://github.com/legend-exp/LegendDataManagement.jl/pull/180](url) first

theHenks:
Closed in favour of #199

PR #182, opened 2026-06-26
read_ldata: per-detector slicing on event tiers (jlevt / jlskm / jlpmt)

jonasschlegel:
## What didn't work before
Event-tier files (`jlevt`, `jlskm`, `jlpmt`) store **one row per event**, with every
detector's data packed together in that row. To get a *single* detector's value you
had to load the **entire** event table (all detectors, ~80 columns) and then dig
into the nested `geds` / `spms` / `aux` structure by hand. Inside, the per-detector
values sit in variable-length lists (one list per event), so you also had to find
that detector's slot in each event's list yourself. And the per-detector handling
was hardcoded to germanium (`geds.trig_e_*`) — it didn't work for SiPMs or PMTs.

Event-level / LAr-veto analysis needs per-detector quantities straight from the
event tier (a GED's calibrated energy per event, a SiPM's PE sum, …). Doing it by
hand is slow (you materialize the whole multi-GB table) and error-prone (you index
the per-event lists yourself), and it only worked for GEDs.

- `read_ldata((:geds_e_cusp_ctc_cal, …), l200, (:jlevt, fk, det))` returns a **flat**
  table with simple names (`geds_e_cusp_ctc_cal`, `aux_pulser_aux_trig`, …), already
  **sliced to that detector** — one value per event, not a list.
- Works for **GEDs, SiPMs and PMTs** (the system is detected automatically), not just
  germanium.
- Only the columns you ask for are read from disk, **even when you also pass a
  `filterby`** — so you don't pay for loading the whole table.
- Reading an event tier **without** a detector still returns the original nested view,
  so existing code keeps working.

**Under the hood:** it flattens the nested event group into a `name → path` map,
finds the detector's slot per event via `trig_e_det` (GEDs) or `detector`
(SiPMs/PMTs), and pulls that slot out of each per-event list. Per-trigger vs.
per-detector columns are told apart by name **and** length, so you always get the
right value; an old channel-based file now fails with a clear message instead of
returning wrong numbers.

## Stack
On top of **PR1** (DetectorId-only API) and **PR2** (path resolver / read helper).
Required by **PR4** (cross-tier `filtertier`).

### DIFF
--- a/ext/LegendDataManagementLegendHDF5IOExt.jl
+++ b/ext/LegendDataManagementLegendHDF5IOExt.jl
@@ -119,8 +119,18 @@ function __init__()
     (@isdefined DataPartition) && extend_datatype_dict(DataPartition, "datapartition")
 end
 
-const _evt_tiers = DataTier.([:jlevt, :jlskm])
+const _evt_tiers = DataTier.([:jlevt, :jlskm, :jlpmt])
 const _perdet_tiers = DataTier.([:jlpeaks, :jlhit, :jlpls])
+
+function _evt_persubdet_path(tier::DataTier, sys::Symbol)
+    if tier == DataTier(:jlevt) || tier == DataTier(:jlskm)
+        sys == :geds && return "$tier/geds"
+        sys == :spms && return "$tier/spms"
+    elseif tier == DataTier(:jlpmt)
+        sys == :pmts && return "$tier"
+    end
+    return nothing
+end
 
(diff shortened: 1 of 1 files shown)


jonasschlegel:
@theHenks @DaGeibl ready to review, but we need to merge [https://github.com/legend-exp/LegendDataManagement.jl/pull/181](https://github.com/legend-exp/LegendDataManagement.jl/pull/url) first

theHenks:
Closed in favour of #199

PR #183, opened 2026-06-26
read_ldata: cross-tier filtering via the filtertier kwarg

jonasschlegel:
## What didn't work before
There was no way to filter a read of one tier by a condition that lives in a
*different* tier. Example: "give me this detector's raw waveforms, but only for the
events that pass a quality cut and an energy threshold" — where the cut/threshold are
computed in the `jlevt` (event) tier, not in `raw`. You had to read both tiers
yourself and line them up by hand — and they **don't** line up row-by-row: the event
tier stores, per detector per event, the **row number** in `raw`/`jldsp` where that
event's data sits.

The natural workflow is "pick interesting events at the event level, then pull their
raw waveforms / DSP values for a detector." Without this you load everything and
align indices manually — slow and easy to get wrong, because the event tier indexes
into `raw`/`jldsp` by a per-detector event number, not by row position.

- New `filtertier` kwarg:
  `read_ldata(…, (:raw, fk, det); filterby = <event-level cut>, filtertier = :jlevt)`
  returns only the `raw` rows of that detector for events that pass the cut.
- Works for phy (filter via `jlevt`) and cal (filter via `jldsp`), and symmetrically
  between the sibling per-trigger tiers (`raw` ↔ `jldsp`).
- Still only loads the columns you actually need.

**Under the hood:**
- Event-tier filter: it reads the detector's per-event index column (`*_detevtno`,
  which points at the matching `raw`/`jldsp` row) for the events passing `filterby`,
  then takes exactly those rows from the target tier. (`*_detevtno` is the 1-based
  per-(filekey, detector) row index — unlike `*_dataidx`, which is run-wide in some
  datasets.)
- Sibling per-trigger filter (`raw` ↔ `jldsp`): these are 1:1 row-aligned per
  detector, so it builds a boolean mask from `filterby` and applies it.
- Guards (clear errors instead of silent wrong results): rejects event↔event combos,
  bounds-checks the indices, asserts the 1:1 alignment, and respects `ignore_missing`.

## Stack
Top of the `read_ldata` stack: builds on **PR1** (DetectorId-only API), **PR2** (path
resolver / read helper) and **PR3** (per-detector event-tier slicing). Nothing builds
on it.

### DIFF
--- a/ext/LegendDataManagementLegendHDF5IOExt.jl
+++ b/ext/LegendDataManagementLegendHDF5IOExt.jl
@@ -122,6 +122,13 @@ end
 const _evt_tiers = DataTier.([:jlevt, :jlskm, :jlpmt])
 const _perdet_tiers = DataTier.([:jlpeaks, :jlhit, :jlpls])
 
+# Per-(fk, det) row index into raw/jldsp; *_dataidx can be run-wide so we use *_detevtno.
+const _evt_idx_col = Dict{Symbol,Symbol}(
+    :geds => :geds_detevtno,
+    :spms => :spms_detevtno,
+    :pmts => :detevtno,
+)
+
 function _evt_persubdet_path(tier::DataTier, sys::Symbol)
     if tier == DataTier(:jlevt) || tier == DataTier(:jlskm)
         sys == :geds && return "$tier/geds"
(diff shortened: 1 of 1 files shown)


jonasschlegel:
@theHenks @DaGeibl ready to review, but we need to merge [https://github.com/legend-exp/LegendDataManagement.jl/pull/182](https://github.com/legend-exp/LegendDataManagement.jl/pull/url) first

theHenks:
Closed in favour of #199

PR #185, opened 2026-07-02
Update SolidStateDetectors version to 0.11.4

tdixon97:
This is needed to introduce the new changes by @hervasa2 introduced in SSD 0.11.4,

In general since patches cannot contain breaking changes is there really a reason to pin the exact version and not eg. 0.11?


theHenks:
I am not a SSD expert with respect to the versions. We could also do 0.11 instead? But this looks to me like you need a specific version?

theHenks:
@tdixon97 Should I merge or change to `0.11`?

CC @hervasa2

hervasa2:
> @tdixon97 Should I merge or change to `0.11`?
> 
> 
> 
> CC @hervasa2 

Merge. There is a bug in ssd <0.11.3

